PR #2, opened 2019-09-17
Some updates

Noviny:
I think I've made things better but feel free to overrule any/all things

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ This action for [Changesets](https://github.com/atlassian/changesets) creates a 
 Create a file at `.github/workflows/version.yml` with the following content.
 
 ```yml
-name: Version
+name: Release
 
 on:
   push:
@@ -18,7 +18,7 @@ on:
 
 jobs:
   release:
-    name: Version
+    name: Release
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ This action for [Changesets](https://github.com/atlassian/changesets) creates a 
 
 ### Without Publishing
 
-Create a file at `.github/workflows/version.yml` with the following content.
+Create a file at `.github/workflows/release.yml` with the following content.
 
 ```yml
 name: Release
--- a/README.md
+++ b/README.md
@@ -75,8 +75,24 @@ jobs:
       - name: Create Release Pull Request or Publish to npm
         uses: changesets/action@master
         with:
-          publish: yarn changeset publish
+          publish: yarn release
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
 ```
+
+
+Doing a publish is triggered by the existence of:
+
+```yaml
+        with:
+          publish: yarn release
+```
+
+This command will run `yarn release` in your repository on merging code into master. You will likely want a to add `release` command to your `package.json` that looks like:
+
+```sh
+"release": "yarn build && changeset publish"
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -75,9 +75,7 @@ jobs:
       - name: Create Release Pull Request or Publish to npm
         uses: changesets/action@master
         with:
-          # The --otp=1 is to get around a bug in changesets
-          # It will be fixed in Changesets 2
-          publish: yarn release --otp=1
+          publish: yarn changeset publish
```
mitchellhamilton: The original reasoning behind the publish option was to allow building before publish but this made me think more, I'm thinking this option should be removed and publish should happen if an NPM_TOKEN is provided and a build option should be added that's run before publish. Any thoughts on this?

README.md
```diff
@@ -6,10 +6,10 @@ This action for [Changesets](https://github.com/atlassian/changesets) creates a
 
 ### Without Publishing
 
-Create a file at `.github/workflows/release.yml` with the following content.
+Create a file at `.github/workflows/version.yml` with the following content.
 
 ```yml
-name: Release
+name: Version
```
mitchellhamilton: My thoughts were release is the right word because it's the combination of version + sometimes build + publish (https://github.com/atlassian/changesets/pull/158) any objections to that?

Noviny: I changed the one that was specifically 'copy this if you just want versioning' as opposed to the 'this one does versioning and commit' - to try and make the 'version' vs 'release' definition more consistent. Happy to move it back.

README.md
Noviny: So was the idea it would run `yarn release` which could be their own aliased yarn command doing whatever? But it assumes that they added `changeset publish` somewhere in the release because we pass in the otp?

Agree that it should run a build thing. I wouldn't mind explicitly setting a publish boolean (or setting a publish command being the trigger for this) to make when the bot will publish more explicit than implicit.

Also, since we don't need to pass anything to the changeset command, it might be worth having it run `yarn pre-publish` and then the bot run the changeset command? It cuts the bot off from random publish scripts in non-changeset repos but otherwise seems fine.

Noviny: > setting a publish command being the trigger for this

checked code to confirm this is what it does, will document loudly that the existence of this triggers this.

mitchellhamilton: > So was the idea it would run yarn release which could be their own aliased yarn command doing whatever? But it assumes that they added changeset publish somewhere in the release because we pass in the otp?

Yes, that's correct. I know it's bad, it was a "this is a bad idea but I want this to work and doing this will make it work" thing.

> Agree that it should run a build thing. I wouldn't mind explicitly setting a publish boolean (or setting a publish command being the trigger for this) to make when the bot will publish more explicit than implicit.

SGTM

> Also, since we don't need to pass anything to the changeset command, it might be worth having it run yarn pre-publish and then the bot run the changeset command? It cuts the bot off from random publish scripts in non-changeset repos but otherwise seems fine.

As in make this action run a pre-publish script or the Changesets CLI?

README.md
mitchellhamilton: Hmmm, that makes sense. My thoughts on naming the things release was that I was planning on renaming this action to `changesets/release-action` so having some of the things named version and some release would be strange.

Maybe we could have a version action and a publish action but separating the things because of naming seems like not the best idea?

Noviny: I think sticking with `Release` is simply going to be easier (I don't want to make this two actions if we can help it)

---

meta-discussion: Is it better to have one `changesets` GH action that's configurable with multiple actions, vs having different things for different tasks. The way this was initially set up, I thought we were going to be getting one package, but we've since divided. What's the reasons?

Noviny:
Closing as old

PR #7, opened 2019-11-08
Change PR title in prerelease mode to include tag name

Andarist:
We could match commit message to this - but I don't care about it that much. The PR title is somewhat important as we have in emotion 2 PRs with the same titles right now. If you feel like changing commit message as well then let me know and I'll adjust this or prepare a new PR.


Andarist:
Yeah, I can change this - no problem. Just having hard time to see when this would actually matter (or is the request just about being more future-proof)?

mitchellhamilton:
One case is when pre-releases are on the default branch. Mentioning the branch is kinda unnecessary.

Andarist:
Ok, the use case makes sense (prereleases on a default branch). I think though that in such a case it would still be beneficial to mention the prerelease tag in a title - only the source of this information has to be different than what I've used right now. I'll make this change in a moment - if you think that it doesn't make sense I can adjust later to what you have proposed.

PR #8, opened 2019-11-09
Remove updating pull request title to allow manual editing

Andarist:
I actually haven't included this on purpose. It seems like if one wishes to edit PR's title manually after it gets created then we shouldn't update it back.

This is a minor thing, probably nearly noone will edit those - but I haven't found a strong argument to "force" the title to the automated one, so it seems like this can just be removed.

Feel free to decline the PR though.


mitchellhamilton:
The main case I was thinking of is when the title would change is when a branch enters or exits pre mode, the title should be updated.

PR #16, opened 2020-01-23
Include a link back to the README for folks to read

orta:
Mainly because I had to DDG for the action name to get back to it


Noviny:
Thanks for this!

PR #20, opened 2020-01-30
Bring back former PR title

Andarist:
IMHO current commit msg as a PR title is way worse than a previous version, I don't care about commit msg following some convention, but it would be great to keep a PR title nice & clean.

If this is something you don't want to merge then this really should be configurable.

cc @Noviny @harrysolovay


Andarist:
Should we allow for configuring commit message & pr title separately? Should both be put inside changesets JSON config? Is it OK to jump on this back & forth within a single major of changesets (I mean - the default has just been changed and we are considering changing it back now)?

harrysolovay:
@Andarist @mitchellhamilton apologies if I introduced undesirable behavior with that PR. Following Conventional Commits was a short-term solution to the problem of commit-message configurability. If you two can hold off on merging this, I can work on another PR this weekend. Because there's no definitive template string syntax for JSON, it might be best if––from the JSON config––we reference a script which exports a commit-message-forming function:

```js
module.exports = (type, scope, description) =>
  `${type}(${scope}): ${description}`;
```

This way we can package and publish each style's commit-message creator, and reference them relative to `node_modules`.

```diff
{
  "$schema": "https://unpkg.com/@changesets/config@0.3.0/schema.json",
  "changelog": "@changesets/cli/changelog",
  "commit": true,
+ “messages”: {
+   "commits": “changesets-conventional-commits”,
+   "pull-request": "..."
+ },
  "linked": [],
  "access": "restricted",
  "baseBranch": "master"
}
```

^ The same would be true for PR title. Please let me know if you two agree with this approach.

Noviny:
I’m actually pretty happy to merge this and then make the commit message configurable and add no config for the bot PR.

Still agree with the above config for adding a way to configure the git commits. The two messages you need though are for `add` and `version` (the two actions that can lead to commits) - this would then preferably use the `version` commit function.

—

Re: changing this and then changing back within majors - we try not to own commit message formatting as API, so I’m not too worried. Would be likely to make it a breaking change once we’ve got deliberate convention

harrysolovay:
@Noviny gotcha––just to be certain we're on the same page: you agree with the mechanism I mocked out above, with the exception of changing the naming to `add` / `version`?

Noviny:
Yep, that seems fine to me. I think I'd like to change the defaults back to something that doesn't line up with conventional commits (so there's no implicit link there - if you are using changesets you probably don't need conventional commits as that's two versioning strategies)

Want to add this to the other issue before you pick it up?

harrysolovay:
@Noviny agreed––no need to have conventional commits as a default. I'll give this a whirl.

harrysolovay:
This is discouraging: I've cloned `atlassian/changesets`, and the `watch` script exists with this:

Screen Shot 2020-02-02 at 1 15 56 PM

It's not a linking problem. The TS does seem to get compiled. Why is it that there's no `types` field nor declaration files being generated for these packages? They do seem to depend on one another's types. I feel like I'm missing something here. Can you please explain this workflow a bit? A `CONTRIBUTING.md` in this action and the `atlassion/changesets` repos would be greatly appreciated!

Andarist:
@harrysolovay there is a chance that recently merged upgrade to TS has broken this. I think I've fixed it with https://github.com/atlassian/changesets/pull/271 .

harrysolovay:
@Andarist I've managed to get it synced with the upstream & building, but something's still wonky with the configuration. Even though TS 3.7.4 is being used, control flow / type guard info isn't being picked up properly––I'm trying to add a new condition to the `parse` fn in `packages/config/src/index.ts`. This new condition checks for the presence of `json.messageFormatters` (which I've added in the `types` package). If `json.messageFormatters` exists, it proceeds to check for the presence of child props `add` and `version`.  I'm getting the following:

Screen Shot 2020-02-02 at 2 52 27 PM

Screen Shot 2020-02-02 at 2 52 37 PM

Also, optional chaining is broken. I've made sure that the `yarn.lock` and `node_modules` are fresh, incase it was an outdated version of TypeScript... restarted the language service like a mad man. It doesn't seem to be related to version. Any ideas about what's going on here?

Andarist:
The first problem is expected, once you have used forEach you have invalidated TS refinement - it cannot be sure that a function call have not removed a property which you have just checked for existence. 

As to optional chaining - not sure, could you check what version of VSCode are you using and what version of TS is being used in it? The latter you should be able to check in the bottom right corner of the app window. VSCode has this weird behavior that it uses its own TS version over the one that is installed in opened project.

harrysolovay:
Interesting! I wasn't aware of that behavior until now. Cool!

Yea––I'm definitely on the right TS version. I see `@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser` versions as both `^2.6.0`. I don't think these are valid. The current version [seems to be](https://www.npmjs.com/package/@typescript-eslint/parser) 2.18.0... also, the problem stems from not specifying `eslint` `parserOptions`. The following addition fixes optional chaining parse errors.

```diff
+ "parserOptions": {
+   "createDefaultProgram": true,
+   "ecmaVersion": 2020,
+   "project": "tsconfig.json",
+   "sourceType": "module"
+ },
```

I'll add this in with the rest of the changes.

harrysolovay:
@Andarist –– apologies to keep pinging you for assistance with little workflow details. I'm running into a type error again :(

Screen Shot 2020-02-02 at 5 04 12 PM

Any idea what's goin' on?

Andarist:
Have you rebased on the current master? I think its fixed there, if not - please push out your changes and list exact repro steps how to reach this situation from a fresh clone to an er error

harrysolovay:
I didn't rebase. I forked yours & then fetched the upstream. Are you saying I need to rebase your fork's commits with `atlassian/changesets` master? I'm gonna struggle with that.

Instead, can you please keep me posted about `atlassian/changesets` master? I can pick this up again when its development environment is fixed.

Andarist:
I believe master has been fixed. You can note down your commits, checkout on master and cherry-pick your commits

harrysolovay:
@Andarist I'm unable to submit a draft PR for feedback, but [here's the fork](https://github.com/harrysolovay/changesets) off of which I'm working. Quite a few tests are breaking, and I don't have the ramp-up time necessary to fix'em. I was wondering if you'd be willing to take a look at these changes and give feedback?

PR #23, opened 2020-04-14
Add parameter to control version script

luisherranz:
I have added an optional parameter called `version` that allows you to take control of the `changeset version` command. This is useful when you need to add additional logic to the command, like for example updating other files.

I have tested it in one of our repositories and it is working fine:
- The npm script: https://github.com/frontity/wp-plugins/blob/dev/package.json#L25
- The code of the script: https://github.com/frontity/wp-plugins/blob/dev/scripts/version.js
- The action execution: https://github.com/frontity/wp-plugins/runs/586330983?check_suite_focus=true
- The PR created: https://github.com/frontity/wp-plugins/pull/30

As you can see, there's an additional file (`plugins/frontity-headtags/plugin.php`) updated in the PR, generated by our script.


luisherranz:
Thanks to you @mitchellhamilton 🙂

PR #26, opened 2020-04-19
Add outputs variable `published` and `publishedPackages`

### LINKED ISSUE #25: Any way to know if a publish occurred?
I'd like to send a Slack notification whenever a publish happens, but from what I can tell this action doesn't give me any way to determine if it just opened a PR or did a full publish. Maybe I'm missing something? If not, consider this a feature request!

ocavue:
Fixes #25


Noviny:
Hi @chasemccoy this looks good! Thanks for the contribution, and for documenting it.

I was wondering just quickly if there was any value in using the list of published packages here so the output was either:

`false` or `array` so that you can include more info about what happened - I could see that being valuable to be able to send to slack.

If not, that's totally fine, and I'm happy to merge once you get back to me.

chasemccoy:
@Noviny to be clear, I didn’t contribute to this PR, but I think your idea is great! I would definitely see the value in knowing which packages were published. 

@ocavue would you mind adding that? If not I can take a stab at it myself.

Noviny:
*headdesk* Sorry for name typing failure

ocavue:
@Noviny 

That's a good idea to provide more information about a publishing. 

Personally, I don't like a variable with different types. So I added another output field `publishedPackages` to present a array.

Noviny:
Awesome, thanks!

PR #30, opened 2020-05-08
fix: respect .npmrc if already present

### LINKED ISSUE #29: Don't overwrite ~/.npmrc
We publish to a private registry, and we have our `.npmrc` files initialized as needed to access those registries. Please don't blast away our configuration:

https://github.com/changesets/action/blob/2eb58c89dcc7f84b755c18b3cf4a6161787a3197/index.ts#L88-L91

emmenko:
Fixes #29


jmackay-io:
Wow, awesome turnaround time - looks like exactly what we needed in #29 🙏

PR #32, opened 2020-06-07
fix: restore previous branch after pushing to changesets Pull Request

emmenko:
I noticed that after the action runs, the active branch is set to e.g. `changeset-release/master`.

I would expect that the original active branch is restored so that follow up actions assume to run on the correct branch.


mitchellhamilton:
Thanks for the PR but I don't think this is something we should do. People could be relying on being on being on the release branch if the release PR was created/updated(e.g. https://github.com/davidkpiano/xstate/blob/48cde74e75e4c119e12c61f34b9e82ab53268dbe/.github/actions/change-release-intent/main.js, https://github.com/davidkpiano/xstate/blob/48cde74e75e4c119e12c61f34b9e82ab53268dbe/.github/workflows/release.yml) and it's much easier to change back to the previous branch than it is to find out if the release PR was created/updated.

tbh, I think this whole action needs to be restructured but I don't think making changes like this that could break people, especially given that we currently recommend people use `changesets/action@master` makes sense.

emmenko:
I see, thanks for the feedback.

It's not a big deal, I'm just checking out the branch manually now. I guess for me it wasn't clear what the behavior was. Maybe this can be documented a bit better in the README.

But if you plan to refactor/restructure the action anyway, we can keep it as-is for now.

Issue #33, opened 2020-06-09
Incompatible with @changesets/cli v2.8.0


ocavue:
In [@changesets/cli v2.8.0](https://github.com/atlassian/changesets/commit/5dc389f9cc03a59ef44f34243efe8ad00e35d465#diff-268d4deba624b8bba84127cce7721fcdR79-R86), the git tag format for single-package repositories has changed, which make [this regular expression](https://github.com/changesets/action/blob/6f5135b203d5101ecf2ef3eea53c5cc8bfad56c6/index.ts#L110) not longer working.

This causes two problems when using changesets/action with a single-package repositorie:

1. Can not create GitHub Release automatically.
2. The outputs (`published` and `publishedPackages`) from the Action don't contain correct package information.

ocavue:
cc @Andarist

Andarist:
Thanks for raising this! I'll try to fix this asap.

Andarist:
Sorry for the delay - but I have good news. There is a pending PR to fix this: https://github.com/changesets/action/pull/35 - we should be able to merge & release it soon.

PR #34, opened 2020-06-16
Migrate away from deprecated get-workspaces

Andarist:
Even though this is not strictly related, this will help me fix this issue: https://github.com/changesets/action/issues/33 in a followup PR.

It also brings `lerna` support with it as it has been added some time ago to the `@manypkg/get-packages`.


index.ts
```diff
@@ -75,15 +75,8 @@ import readChangesets from "@changesets/read";
     console.log(
       "No changesets found, attempting to publish any unpublished packages to npm"
     );
-    let workspaces = await getWorkspaces({
-      tools: ["yarn", "bolt", "pnpm", "root"]
-    });
-
-    if (!workspaces) {
-      return core.setFailed("Could not find workspaces");
```
Andarist, line 83: `getPackages` always returns something, even if its an empty array (or throws, of course, but that's just handled by the top-level catch here)

PR #37, opened 2020-06-30
Do not throw if CHANGELOG.md doesn't exist

Feiyang1:
Fixes 2 issues:
1. PR is not created if CHANGELOG.md doesn't exist for a changed package. It could happen when the package is ignored, or only devDependencies are updated and the package doesn't have CAHNGELOG.md ( it could be new package, or an internal package that shouldn't have a CHANGELOG.md)

2. We incorrectly display packages that are updated, but don't have changelog entries (means it's not released) in the PR body. It could happen when the package is ignored, or only devDependencies are updated.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -249,6 +249,12 @@ ${
         (
           await Promise.all(
             changedPackages.map(async (pkg) => {
+              
+              const changelogPath = path.join(pkg.dir, "CHANGELOG.md");
+              if (!fs.pathExists(changelogPath)) {
+                return;
+              }
+
               let changelogContents = await fs.readFile(
                 path.join(pkg.dir, "CHANGELOG.md"),
                 "utf8"
@@ -258,6 +264,11 @@ ${
                 changelogContents,
                 pkg.packageJson.version
               );
+
+              if (!entry.content) {
+                return;
+              }
+
               return {
                 highestLevel: entry.highestLevel,
                 private: !!pkg.packageJson.private,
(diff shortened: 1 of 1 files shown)


ifiokjr:
This issue bit me today while trying to release `pre-release` packages. Hopefully, it can be merged soon.

mitchellhamilton:
I think this is resolved by #38? If not, could you fix the conflicts?

Andarist:
> I think this is resolved by #38? 

I don't believe it would be - that PR was only focused on what PR body gets generated.

@Feiyang1 We have now a better test setup so adding a test for this issue would be highly appreciated as well.

Feiyang1:
I think the first problem is resolved as well by #38, as `changeset version` will create a CHANGELOG.md file if it doesn't already exist for a version updated package, and we only read CHANGELOG.md for version updated packages.

Or Could a version updated package not have a CHANGELOG.md file? @Andarist

Andarist:
> Or Could a version updated package not have a CHANGELOG.md file? @Andarist

It should always be updated AFAIK.

---

So to understand this a bit better - there were situations when we've attempted to read changelogs and failed when some files were touched in a package but the package itself was not updated and this has only happened because of the logic in `getChangedPackages` which has been changed with #38 and currently there is no problem as we check if `pkg.version` has changed before classifying a package as updated, right? If that's the case - then yeah, #38 fixed both problems 🎉 

I guess we could a regression test for this scenario, but it's also not a top priority as things are working correctly now and the regression chance is rather slim in this area.

Feiyang1:
Correct. Closing as it is resolved.

PR #38, opened 2020-07-02
Fixed an issue with ignored package being included in PR body incorrectly

Andarist:
PR body included a header for an ignored package that only had its dependency bumped. This was caused by git-dependent logic in [`getChangedPackages`](https://github.com/changesets/action/blob/c240b08dfbd25dccd1f7df75cdf3639e6e65a741/utils.ts#L17) as in this case `ignored-pkg/package.json` was updated but `ignored-pkg/CHANGELOG.md` was left untouched, so there was nothing to add as contents of this created section.

I've decided to refactor this from git-dependent logic to version-dependent one. I just record all versions before `version` script gets executed and compare to versions available after it.

I wanted to cover this case with a test and as I didn't want to write unit tests for the changed implementation details I've decided to set up integration tests to test the workflow as a whole. To be able to do this I had to restructure files a little bit so I could mock some files (like git utils) and so I could just run a particular "command" with given custom inputs (removing a hard dependency on `@actions/core` from those dependencies).


src/run.ts
```diff
@@ -0,0 +1,272 @@
...
+  await gitUtils.switchToMaybeExistingBranch(versionBranch);
+  await gitUtils.reset(github.context.sha);
+
+  let versionsByDirectory = await getVersionsByDirectory(cwd);
+
+  if (script) {
+    let [versionCommand, ...versionArgs] = script.split(/\s+/);
+    await exec(versionCommand, versionArgs, { cwd });
+  } else {
+    // technically
```
mitchellhamilton, line 175: ?

Andarist: not sure what I have wanted to write here 😅 gonna remove this in a moment

PR #59, opened 2020-10-28
Update index.ts

### LINKED ISSUE #58: project's .npmrc not respected
Right now if you have an `.npmrc` at the project root that is not respected but rather the following piece of code just checks the `.npmrc` in the HOME directory

https://github.com/changesets/action/blob/cd904b92c71a9cde409529b9eb7bf7fd05ea5858/src/index.ts#L44-L53

Even the ideal flow for npm or any process is they check for `.npmrc` at the project root if not then they look for `.npmrc` in the HOME directory.

kamleshchandnani:
Check for `npmrc` in project root first then check in HOME directory. If it's not found at either place then create one in the the HOME directory.

fixes #58

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -41,13 +41,14 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
         "No changesets found, attempting to publish any unpublished packages to npm"
       );
 
-      let npmrcPath = `${process.env.HOME}/.npmrc`;
-      if (fs.existsSync(npmrcPath)) {
+      let npmrcHomePath = `${process.env.HOME}/.npmrc`;
+      let npmrcProjectRootPath = `${process.env.PWD}/.npmrc`;
+      if (fs.existsSync(npmrcProjectRootPath) || fs.existsSync(npmrcHomePath)) {
         console.log("Found existing .npmrc file");
       } else {
         console.log("No .npmrc file found, creating one");
         fs.writeFileSync(
-          npmrcPath,
+          npmrcProjectRootPath,
           `//registry.npmjs.org/:_authToken=${process.env.NPM_TOKEN}`
         );
       }
--- a/src/index.ts
+++ b/src/index.ts
@@ -48,7 +48,7 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
       } else {
         console.log("No .npmrc file found, creating one");
         fs.writeFileSync(
-          npmrcProjectRootPath,
+          npmrcHomePath,
           `//registry.npmjs.org/:_authToken=${process.env.NPM_TOKEN}`
         );
       }


src/index.ts
```diff
@@ -41,13 +41,14 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
         "No changesets found, attempting to publish any unpublished packages to npm"
       );
 
-      let npmrcPath = `${process.env.HOME}/.npmrc`;
-      if (fs.existsSync(npmrcPath)) {
+      let npmrcHomePath = `${process.env.HOME}/.npmrc`;
+      let npmrcProjectRootPath = `${process.env.PWD}/.npmrc`;
```
Andarist, line 45: is there any particular reason why you have chosen this over `process.cwd()`?

Andarist (commented):
LGTM in general - but I have not been using `.npmrc` files much so I'm not super confident myself with merging this, would prefer if @mitchellhamilton would take a look at this first

mitchellhamilton:
I'm not sure why we would want to do this?

Creating the `.npmrc` in the current working directory is definitely unsafe since since we commit from there and most people probably won't have `.npmrc` in their gitignore and people frequently have `.npmrc` committed in projects that set a certain registry for a scope but have their authentication set globally for example.

For the linked issue, won't npm prefer any settings set in the local `.npmrc` over the global `.npmrc` so writing the global `.npmrc` shouldn't cause any issues?

src/index.ts
kamleshchandnani: `process.env.PWD` is a safer way compared to `process.env.cwd()` because while your process is running and if one of the commands in that process changes the working directory while it's still executing the value of `process.env.cwd()` will change at all the places where it's being referred. `process.env.PWD` guarantees that your current working directory will be set to a path when the command actually started executing and wouldn't change during the lifecycle of the process

kamleshchandnani:
> I'm not sure why we would want to do this?
> 
> Creating the `.npmrc` in the current working directory is definitely unsafe since since we commit from there and most people probably won't have `.npmrc` in their gitignore and people frequently have `.npmrc` committed in projects that set a certain registry for a scope but have their authentication set globally for example.
> 
> For the linked issue, won't npm prefer any settings set in the local `.npmrc` over the global `.npmrc` so writing the global `.npmrc` shouldn't cause any issues?

Sure makes sense. I'll make changes and write the `.npmrc` to the Home directory.

mitchellhamilton:
To be clear, I'm unsure what problem this PR solves since npm should respect config in a local .npmrc over the global .npmrc + not writing the global config if a local config exists could cause problems if people have a local .npmrc with some config committed to their repo(but without auth)

kamleshchandnani:
> To be clear, I'm unsure what problem this PR solves since npm should respect config in a local .npmrc over the global .npmrc + not writing the global config if a local config exists could cause problems if people have a local .npmrc with some config committed to their repo(but without auth)

@mitchellhamilton I've an `.npmrc` at the project root but still since changeset checks for `.npmrc` in home directory and if it doesn't exists it just creates one and when it publishes it's referring to the global one and I get failure because my registry is different but the project's `.npmrc` is not getting picked up and I need to unnecessarily write this extra step in my action.

```yaml
- name: Creating .npmrc
  run: |
    cat << EOF > "$HOME/.npmrc"
      email=my@email.com
      //registry.npmjs.org/:_authToken=$NPM_TOKEN
    EOF
  env:
```

mitchellhamilton:
> the project's .npmrc is not getting picked up

Hmmmm, I think that'll be because of https://github.com/atlassian/changesets/pull/448. I don't think changing what's being changed in this PR would fix that.

kamleshchandnani:
Oh okay alright! maybe I'll close this then.

Issue #66, opened 2021-01-19
Fails when using Yarn v2 + PnP


jdb8:
It looks like there's a line that explicitly references `node_modules` here: https://github.com/changesets/action/blob/master/src/run.ts#L186. When using Yarn PnP, this fails, since node_modules doesn't exist.

```
Error: Cannot find module '/home/runner/work/<pkg>/<pkg>/node_modules/@changesets/cli/package.json'
Require stack:
- /home/runner/work/_actions/changesets/action/master/dist/index.js
Require stack:
- /home/runner/work/_actions/changesets/action/master/dist/index.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:797:15)
    at Function.Module._load (internal/modules/cjs/loader.js:690:27)
    at Module.require (internal/modules/cjs/loader.js:852:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at f (/home/runner/work/_actions/changesets/action/master/dist/index.js:1:293)
    at f (/home/runner/work/_actions/changesets/action/master/dist/index.js:1:233)
    at p (/home/runner/work/_actions/changesets/action/master/dist/index.js:1:544)
    at Object.w [as runVersion] (/home/runner/work/_actions/changesets/action/master/dist/index.js:906:3310)
    at async /home/runner/work/_actions/changesets/action/master/dist/index.js:908:1597 {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/home/runner/work/_actions/changesets/action/master/dist/index.js'
  ]
}
```

Andarist:
Does `require.resolve` work with PnP in a way that we could use the resolved path to read the `package.json` file using `fs` APIs?

merceyz:
Yes, you use `require.resolve` just like you normally would

```js
const path = require('path');

const cliPath = require.resolve('@changesets/cli/package.json', { paths: [cwd] });

const changesetsCliPkgJson = require(cliPath); // or JSON.parse(fs.read*(cliPath) )

await exec('node', [path.resolve(cliPath, '../cli/bin.js'), cmd], { cwd });
```

While you should always use something like this and not make assumptions about the `node_modules` layout, since this is an action and thus running outside of the project you'll need to require the `.pnp.(c)js` file and call its `setup` method to have PnP enabled

Andarist:
Would you like to prepare a PR for this?

jacobwgillespie:
@Andarist I've opened a PR for this here: #67

[...]

PR #67, opened 2021-01-28
Resolve @changesets/cli with resolve-from

jacobwgillespie:
This switches `run.ts` from assuming the `node_modules` directory location to using `require.resolve` under the hood to locate the real location of `@changesets/cli`. 

It uses the [resolve-from](https://github.com/sindresorhus/resolve-from) module to do the resolving from the current directory, otherwise just a plain `require.resolve` would attempt to locate the `@changesets/cli` nearest to `run.ts`, rather than the user's project.

As mentioned by @merceyz in https://github.com/changesets/action/issues/66#issuecomment-763501827, you can pair this with `NODE_OPTIONS="-r <cwd>./.pnp.(c)js"` to enable support for Yarn 2 / PnP.

Related to #66


merceyz:
> It uses the resolve-from module to do the resolving from the current directory, otherwise just a plain require.resolve would attempt to locate the @changesets/cli nearest to run.ts, rather than the user's project.

You can use the `paths` option, no need to grab a npm package for this
https://nodejs.org/api/modules.html#modules_require_resolve_request_options

jacobwgillespie:
> You can use the `paths` option as I mentioned in [#66 (comment)](https://github.com/changesets/action/issues/66#issuecomment-763501827), no need to grab a npm package for this
> [nodejs.org/api/modules.html#modules_require_resolve_request_options](https://nodejs.org/api/modules.html#modules_require_resolve_request_options)

Unless I'm misunderstanding, it's not _quite_ that simple, as `paths` entirely replaces the default resolution logic's lookup directories, so specifying `paths: [cwd]` would only allow the module to be present at `$PWD/node_modules/x`, but `$PWD/../node_modules/x` would not resolve, etc.

There's a way to build a list of directories that should be checked and pass that into `paths`, as well as a few edge-cases around resolving directory symlinks to handle, but that that point you've ended up recreating the body of `resolve-from`, minus the tests: https://github.com/sindresorhus/resolve-from/blob/main/index.js

merceyz:
> Unless I'm misunderstanding, it's not quite that simple, as paths entirely replaces the default resolution logic's lookup directories, so specifying paths: [cwd] would only allow the module to be present at $PWD/node_modules/x, but $PWD/../node_modules/x would not resolve, etc.

As the docs specify `paths` is where it should start seaching, parent folders are also checked

> Each of these paths is used as a starting point for the module resolution algorithm, meaning that the node_modules hierarchy is checked from this location.

jacobwgillespie:
> As the docs specify `paths` is where it should **start** searching so parent folders are also checked

Oh nice, yes I was indeed misunderstanding. I've updated to use `require.resolve` directly 👍

Andarist:
> As mentioned by @merceyz in #66 (comment), you can pair this with NODE_OPTIONS="-r <cwd>./.pnp.(c)js" to enable support for Yarn 2 / PnP.

Is this worth documenting?

jacobwgillespie:
> Is this worth documenting?

Probably yeah, I updated the PR with a README section. The `.pnp.js` file needs to be required with an absolute path, not a relative path, to ensure it's resolved even when scripts are run from subdirectories, so it's likely good to have a README example.

Andarist (approved):
I will let it sit for a couple of days - maybe someone else would like to take a look at this before merging as well. If no one will I will merge this soon(-ish) - in case I forget about it feel free to ping me.

merceyz (approved):
LGTM - Might want to update the title and description though

jacobwgillespie:
Okay, pinging @Andarist, this is still working for us (feel free to wait longer if you'd like, just following up on the reminder request).

mitchellhamilton (commented):
tbh, wouldn't it be easier to tell people to do this instead (the code change seems fine but for the documentation, the below seems easier)

```yaml
with:
  version: yarn changeset version
```

jacobwgillespie:
> wouldn't it be easier to tell people to do this instead

Yep, that does seem simpler - I've updated the PR to recommend a custom version command ✅

jacobwgillespie:
This PR breaks the action for non-Yarn-2 projects it seems - I've opened PR #71 to fix.

PR #71, opened 2021-02-15
Fix failure to resolve changeset CLI in GitHub Actions

jacobwgillespie:
PR #67 broke the action for me for projects that _don't_ use Yarn 2 / PnP. It appears that the issue is that in an Actions environment, the root workspace directory is not a real directory, so it causes issues with a plain `require.resolve`. When testing the action locally, I didn't catch this error since the workspace directory wasn't a symlink.

This PR uses [`resolve-from`](https://github.com/sindresorhus/resolve-from) to locate the changeset CLI, like the other PR originally did, as that module has code to work around symlinked directories.

I can confirm that this branch works for my non-Yarn-PnP projects.


src/run.ts
```diff
@@ -181,20 +182,16 @@ export async function runVersion({
     let [versionCommand, ...versionArgs] = script.split(/\s+/);
     await exec(versionCommand, versionArgs, { cwd });
   } else {
-    let changesetsCliPkgJson = await require(require.resolve(
-      "@changesets/cli/package.json",
-      { paths: [cwd] }
```
merceyz, line 186: Why not just call `fs.realpathSync` just like it does and avoid the dependency?
https://github.com/sindresorhus/resolve-from/blob/4f02820afe9b4e9c379f4bb2745debdb49087d4f/index.js#L16

jacobwgillespie: Same reason as mentioned in #67: tests and robustness. This module is small and is well-used and well tested in the npm ecosystem (it's already present in `yarn.lock` here, for instance). Also the full code from `resolve-from` is:

```javascript
let fromDirectory = cwd
try {
  fromDirectory = fs.realpathSync(fromDirectory)
} catch (error) {
  if (error.code === 'ENOENT') {
    fromDirectory = path.resolve(fromDirectory)
  } else {
    throw error
  }
}
```

Andarist (approved):

PR #80, opened 2021-03-22
Feature: Comment on released PRs and issues

ayshiff:
Closes [#511](https://github.com/atlassian/changesets/issues/511)

As seen with @mitchellhamilton, this PR is a duplicate of [#25](https://github.com/changesets/bot/pull/25) and has been moved here to make the feature more configurable and consistent.

This PR adds a new feature that allows to comment on PRs and issues that have been released.
It also adds a new input `comment` which allows to activate or not the feature.

My work is inspired by what [semantic-release/github](https://github.com/semantic-release/github) has done.

Here is a schema showing the workflow to retrieve the released Pull Requests and Issues.

<p align="center">

</p>

## Comment message

Here is a comment message example:

###  🦋  This work has been released in release version: v0.4.3

Release link: https://github.com/backstage/backstage/releases/tag/v0.4.3

Note that the message is the same for PRs and issues.   
It might be interesting to have two different messages for the different cases?

## Example

I also created a [codesanbox](https://codesandbox.io/s/winter-bird-pwn6v?file=/src/index.js) where you can see the logic  to get the issues and pull-requests associated to a release more easily. 
**NOTE**: In the codesanbox, considering that the code does not run at the time the release is created, we will get more recent commits. This will not happen in the real world.

## Notes

I couldn't find a better way to find the tag associated to the release than looking through the tags until I found the one corresponding to the release.


ayshiff:
I will change the place where we create comments to avoid putting it in `createRelease()`.
We will create the comments only once, after retrieving all the released issues and PRs for all the packages and avoid duplicates.

mitchellhamilton:
Could you create a repo and use your fork of the action on it and test it out so we can see what it looks like?

src/run.ts
```diff
@@ -108,6 +260,7 @@ export async function runPublish({
         createRelease(octokit, {
           pkg,
           tagName: `${pkg.packageJson.name}@${pkg.packageJson.version}`,
+          comment: false
```
Andarist, line 263: why this is hardcoded to false? 🤔 shouldn't we use a configured value here?

src/run.ts
```diff
@@ -10,14 +10,16 @@ import {
   getChangedPackages,
   sortTheThings,
   getVersionsByDirectory,
+  getReleaseMessage,
 } from "./utils";
 import * as gitUtils from "./gitUtils";
 import readChangesetState from "./readChangesetState";
 import resolveFrom from "resolve-from";
+import issueParser from "issue-parser";
```
Andarist, line 18: q: in the GitHub interface we have "Linked issues", I assume that this list is received by their frontend from their backend and that they don't have to prepare the PR body to get that. Would it be possible to leverage that? Or are there some limitations regarding this technique? One thing that comes to my mind - the list might include more than we want but maybe it's associated with some metadata that we could leverage to filter that list?

src/run.ts
```diff
@@ -47,10 +52,156 @@ const createRelease = async (
...
+    
+        1. Retrieve the tag associated with the release
+        2. Take the commit sha associated with the tag
+        3. Retrieve all the commits starting from the tag commit sha
+        4. Retrieve the PRs with commits sha matching the release commits
+        5. Map through the list of commits and the list of PRs to
+           find commit message or PRs body that closes an issue and
+           get the issue number.
+        6. Create a comment for each issue and PR
+      */
```
Andarist, line 85: Each released thing should be associated with a changeset. I know that it's still possible to just release something while piggybacking on the already prepared release and close something that way but... should we care about it? Shouldn't we guide users to the best practices?

What if we'd make this changeset-based and just appropriately mix that info with additional data (like the link to the release)? We could also add contents of the relevant changesets to the generated comment.

src/run.ts
ayshiff: This is hardcoded to false because we want to comment on whole repository release rather than on each package release.

ayshiff:
> Could you create a repo and use your fork of the action on it and test it out so we can see what it looks like?

Here is an example: 

- Issue comment: https://github.com/ayshiff/changesets-action-test/issues/1#issuecomment-805863301
- PR comment: https://github.com/ayshiff/changesets-action-test/pull/2

src/run.ts
ayshiff: I agree with you, I wish I had found a more appropriate solution but I think we don't have the necessary information in the API response (see https://docs.github.com/en/rest/reference/pulls#get-a-pull-request).

I tried with a Pull Request that has several related issues and they don't appear in the API response.

src/run.ts
Andarist: Hm, there is no such thing as "whole repository release" if I recall correctly. The other `createRelease` is for single-package repositories but in monorepos we are just looping through all released packages so there is 1 to 1 relation between released packages and created releases (no extra repository-wide release)

ayshiff: Oh right I misunderstood what @mitchellhamilton meant here https://github.com/changesets/bot/pull/25#pullrequestreview-618021405

I'll use the `comment` value from the input property.

src/run.ts
Andarist: There seems to be a way using the GraphQL API: https://github.community/t/get-all-issues-linked-to-a-pull-request/14653/6

src/run.ts
```diff
@@ -47,10 +54,155 @@ const createRelease = async (
...
+  await Promise.all(
+    [...new Set([...pulls, ...issues].map(({ number }) => number))].map(
+      async (number) => {
+        const issueComment = {
+          ...repo,
+          issue_number: number,
+          body: getReleaseMessage(htmlUrl, tagName),
+        };
+
+        octokit.issues.createComment(issueComment);
```
Andarist, line 195:
```suggestion
        return octokit.issues.createComment(issueComment);
```

Otherwise the outer `Promise.all` won't wait on those calls.

src/run.ts
Andarist: This doc also claims that it's possible to query that ingo using filters: https://docs.github.com/en/github/searching-for-information-on-github/searching-issues-and-pull-requests#search-for-linked-issues-and-pull-requests

But I couldn't make it work to get actual issue numbers.

mitchellhamilton:
Could you make your test repo have a bunch of packages, I'm curious to see what that looks like since there's no single release on GH to link to?

Andarist:
@mitchellhamilton since creating those comments is done currently just after creating a release then this has to "spam" issues, creating a comment for each release from that single run. This definitely needs improvement and we should aggregate stuff to create at most one comment per issue.

ayshiff:
> @mitchellhamilton since creating those comments is done currently just after creating a release then this has to "spam" issues, creating a comment for each release from that single run. This definitely needs improvement and we should aggregate stuff to create at most one comment per issue.

I will try to aggregate the different issues to avoid duplicates and provide more complete comments.
Probably something like this:

```ts
// Inside createReleaseComments()
return [...new Set([...pulls, ...issues].map(({ number }) => number))].map(
  (number) => ({
      ...repo,
      issue_number: number,
      body: getReleaseMessage(htmlUrl, tagName),
    })
)

// In runPublish()
// Aggregate the differents comments
    const comments = await Promise.all(
      releasedPackages.map((pkg) =>
        createRelease(octokit, {
          pkg,
          tagName: `${pkg.packageJson.name}@${pkg.packageJson.version}`,
          comment,
        })
      )
    ).then((values) => 
       // Combine by issue number
    )
```

We will probably have something like this to build the body of the comment:

```json
{
  "<ISSUE_NUMBER>": [
    {
      "htmlUrl": "<HTML_URL>",
      "tagName": "<TAG_NAME>",
    }
  ]
}
```

Tell me if this makes sense to you.

Andarist:
@ayshiff could you take a look at [this comment](https://github.com/changesets/action/pull/80#discussion_r600353122)? I think that answering those questions should be done first as those are high-level questions

src/run.ts
ayshiff: I agree with that!

So the idea would be to do that in the `runVersion()` method when we have changesets.   
In the `readChangesetState()` response we can retrieve the changeset releases:

```ts
// changesets: { id, summary, releases }[]
let { preState, changesets } = await readChangesetState(cwd);
```

We can then base our logic on these changesets releases.

ayshiff:
I tried to answer the different requests in my last commit. (It's still a work in progress)

I added a condition to create comments only when we have a changeset.
I also make sure that there is only one comment per issue and that we can aggregate some information in this comment.

I currently display the entire list of changeset summaries, but I'll try to find a way to display only the information that is relevant to that comment.

I'm a bit busy at the moment but I'll try to find some time to test and improve this.

Andarist:
> I'm a bit busy at the moment but I'll try to find some time to test and improve this.

No problem - I'm a little bit busy right now as well so I might not have time to review this right away. Please ping me when you want me to do the review :)

orta:
( if you want an additional reference, I have built something like this before: https://github.com/danger/peril-settings/blob/master/org/new_tag.ts )

src/utils.ts
```diff
@@ -18,6 +18,21 @@ export async function getVersionsByDirectory(cwd: string) {
   return new Map(packages.map((x) => [x.dir, x.packageJson.version]));
 }
 
+export const getReleaseMessage = (
+  htmlUrls: string[],
+  tagNames: string[],
+  changesetsSummaries: string[]
+) => `###  🦋  This work has been released in release versions: ${tagNames.join(
```
acao, line 25: just a nit, but perhaps you want to either say `version(s)` and use other such possible pluralization shortcuts here, or you could pluralize based on array length

also, what about making this message the default, and allowing a supplied parameter for this line of the issue/PR close comment?

src/run.ts
acao:
```graphql
fragment IssueOrPullRequestData on ReferencedSubject {
  ... on Issue {
    number
    body
    issueState: state
  }
  ... on PullRequest {
    number
    state
  }
}

{
  resource(url: "https://github.com/graphql/graphiql/pull/1914") {
    ... on PullRequest {
      timelineItems(itemTypes: [
        CONNECTED_EVENT, 
        DISCONNECTED_EVENT, 
        CROSS_REFERENCED_EVENT, 
        REFERENCED_EVENT
			], first: 100) {
        nodes {
          __typename
          ... on ReferencedEvent {
            id
            actor {
              url
            }
            isCrossRepository
            isDirectReference
            subject {
              ...IssueOrPullRequestData
            }
          }
          ... on DisconnectedEvent {
            id
            subject {
              ...IssueOrPullRequestData
            }
          }
          ... on CrossReferencedEvent {
            resourcePath
            isCrossRepository
            willCloseTarget
            target {
              ...IssueOrPullRequestData
            }
            source {
              ...IssueOrPullRequestData
            }
          }
        }
      }
    }
  }
}
```

got a little closer, but not quite. using this query, I was able to get the PR referenced from a referenced issue, but not the referenced issue itself :laughing: ...

Andarist: This can now be queried with:
```graphql
{
  resource(url: "https://github.com/graphql/graphiql/pull/1914") {
    ... on PullRequest {
      closingIssuesReferences(first: 99) {
        nodes {
          number
        }
      }
    }
  }
}
```

I will be revamping this PR this week to hopefully land it before the end of the year.

acao:
@ayshiff looking at your last commits, I'm wondering, is this ready? what else is left?

PR #84, opened 2021-04-07
Sending name argument to createRelease octokit call

jonrohan:
This adds the optional `name:` argument to the createRelease call when creating a release. https://octokit.github.io/rest.js/v18#repos-create-release

When this argument isn't provided GitHub makes the decision to use the tagName as the release name, but there's a problem created with webhooks that fire on the creation of a release. When `changesets/action` creates the release we have slack notify of the new releases, but currently it's being output as this: 

image


Andarist:
Do you happen to know for what exact things does it matter? Will this be used as the release's title?
Screenshot 2021-04-07 at 10 59 37

jonrohan:
Yes this becomes the title

PR #90, opened 2021-05-04
fix: print correct log when project npmrc found

### LINKED ISSUE #89: When local `.npmrc` is found the action still says `No .npmrc file found, creating one`
Here's [the repo](https://github.com/kamleshchandnani/changesets-demo) where I have local `.npmrc` which is being considered but the changesets action still says. 
> No `.npmrc` file found, creating one

Here's the [output message from running action](https://github.com/kamleshchandnani/changesets-demo/runs/2493113514?check_suite_focus=true#step:5:16)

I got confused that it isn't picking up local my `.npmrc` even though it is. I think the messaging can be bit more contextual if it finds my local `.npmrc`. So I think we can change the messaging a bit.

kamleshchandnani:
fixes #89

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -43,9 +43,9 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
 
       let npmrcPath = `${process.env.HOME}/.npmrc`;
       if (fs.existsSync(npmrcPath)) {
-        console.log(`Found existing global .npmrc file`);
+        console.log("Found existing global .npmrc file");
       } else {
-        console.log(`No global .npmrc file found, creating one`);
+        console.log("No global .npmrc file found, creating one");
         fs.writeFileSync(
           npmrcPath,
           `//registry.npmjs.org/:_authToken=${process.env.NPM_TOKEN}`
--- a/src/index.ts
+++ b/src/index.ts
@@ -41,13 +41,16 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
         "No changesets found, attempting to publish any unpublished packages to npm"
       );
 
-      let npmrcPath = `${process.env.HOME}/.npmrc`;
-      if (fs.existsSync(npmrcPath)) {
-        console.log("Found existing .npmrc file");
+      let npmrcHomePath = `${process.env.HOME}/.npmrc`;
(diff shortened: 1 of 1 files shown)


mitchellhamilton (commented):
This isn't right, just because there is a .npmrc in the project doesn't mean we don't need to create a global one. People will often have an .npmrc in their project that might configure the cli somehow or set a registry for some scope but it won't have authentication information because it's expected that that will be configured globally. Changing the message to say that we found no _global_ .npmrc would be fine though.

kamleshchandnani:
Sure I've made the changes. But don't you think it'll be better if we atleast somehow notify the user that we are also considering the project `.npmrc`?

Maybe [at this line](https://github.com/changesets/action/pull/90/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R47) we can put an `if(fs.existsSync(projectnpmrcpath)` and put a log and also create the global npmrc file. Something like below. WDYT?

```js
let npmrcHomePath = `${process.env.HOME}/.npmrc`;
let npmrcProjectRootPath = `${process.cwd()}/.npmrc`;

if (fs.existsSync(npmrcProjectRootPath)) {
  console.log(`Found project .npmrc file at ${npmrcProjectRootPath}`);
}

if (fs.existsSync(npmrcHomePath)) {
  console.log(`Found global .npmrc file at ${npmrcHomePath}`);
} else {
  console.log(`No global .npmrc file found, creating one at ${npmrcHomePath}`);
  fs.writeFileSync(
    npmrcHomePath,
    `//registry.npmjs.org/:_authToken=${process.env.NPM_TOKEN}`
  );
}
```

kamleshchandnani:
Is there any update on this?

kamleshchandnani:
@mitchellhamilton does this looks fine?

Andarist:
@kamleshchandnani sorry that this PR didn't get the attention it deserved back in the day. I believe though that this has been, in a way, addressed by some other work and a confusing log is no longer printed out. If you believe that this is still not right - please tell us more and I will make sure to address the issue in a more timely manner this time.

The relevant code lines:
https://github.com/changesets/action/blob/e0c51a52b861b2cc9a389071d2e6bcaf1b5f49bf/src/index.ts#L45-L72

PR #94, opened 2021-05-26
fix: remove git global config flag

dhruvdutt:
Running this action inside a self-hosted GitHub runner that has an Alpine equivalent barebones environment, the changeset action fails since it's trying to set git user & email on global config that looks for `$HOME/.gitconfig`.

Self-hosted instances might also have a shared VM running multiple jobs, updating global git config might pollute the global configs which is not expected.

image

### DIFF
--- a/src/gitUtils.ts
+++ b/src/gitUtils.ts
@@ -4,13 +4,11 @@ import { execWithOutput } from "./utils";
 export const setupUser = async () => {
   await exec("git", [
     "config",
-    "--global",
     "user.name",
     `"github-actions[bot]"`,
   ]);
   await exec("git", [
     "config",
-    "--global",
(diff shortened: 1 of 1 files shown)


dhruvdutt:
@mitchellhamilton @Andarist Can you please review?

dhruvdutt:
@mitchellhamilton @Andarist Can you please take a look at this?

Andarist (approved):
LGTM, @mitchellhamilton do you see any risks associated with this? I couldn't find any

mitchellhamilton (approved):

Andarist:
This got merged in as part of this PR https://github.com/changesets/action/pull/128 because I couldn't push to this fork.

Issue #87, opened 2021-04-23
Skip createRelease


juanpicado:
I'd like to avoid create GitHub releases, I have several packages in non production mode using alpha mode, every time I merge creates a lot of noise on my watchers. I've checked the code and I could not find any option, seems like goes as long publish the package.

I'm using it here https://github.com/verdaccio/verdaccio

It's related with https://github.com/changesets/action/issues/49

Base in your opinion, I'd be willing to work on it.

Andarist:
Do you want to skip releases altogether or do you want to filter them so only for some the releases would be created?

juanpicado:
I'd like be able skip all of them, to make it more simpler.

akphi:
Hi, could we revisit this? I would be willing to help with this as well.

akphi:
@Andarist this should be closed by https://github.com/changesets/action/pull/130

PR #108, opened 2021-08-30
Add a flag to allow skipping Github release creation

akphi:
Fixes https://github.com/changesets/action/issues/87

Added a flag called `SKIP_CREATE_GITHUB_RELEASE` to allow skipping publishing release to Github

@Andarist If you have other suggestions on the config name or the way I pass down the `boolean` please let me know I can fix it ASAP. Would be really appreciative if you could spare some time to look at this. Thank you so much!

Also, I thought about adding some documentation for this on the `README.md` file, but not sure how to organize that, do you have any suggestion here?


Andarist:
@akphi could you describe your use case that requires introducing this option?

akphi:
@Andarist My use case is fairly similar to that of [babel[(https://github.com/babel/babel/tree/main/packages): I have a monorepo with many workspaces/packages. If I were to publish them, I would accumulate hundreds of tags and releases on Github. I would like to use the Github release page a little more sparingly. Ideally, I want to create an accumulated release note per release, but that's off the scope of this feature request.

I could temporarily workaround for now, by setting a post-release script to delete created tags and releases, but it's a hack 😅.

akphi:
Btw, @Andarist I saw your change in https://github.com/changesets/changesets/pull/701 to add `--no-git-tag`. It will help at least not creating the tags, but the releases are still being created by `changesets/action` right?

Andarist:
Would [fixed](https://github.com/changesets/changesets/pull/690) packages and [aggregated releases](https://github.com/changesets/changesets/issues/683) address your needs? Note that I'm planning to introduce those aggregated releases only for linked and fixed packages - at least for now. Introducing that for regular packages is quite problematic when it comes to changeset creation, sorting etc.

> Btw, @Andarist I saw your change in changesets/changesets#701 to add --no-git-tag. It will help at least not creating the tags, but the releases are still being created by changesets/action right?

I think that releases should not be created with this flag (but I have not tested it). It's because releases are created based on the created tags:
https://github.com/changesets/action/blob/37c2edd85bf5f2aa8edb44ce4ea00eb4a55aa9d4/src/run.ts#L88-L114

That being said - the flag that you are asking for is still useful for the use case described in #87. The author of that issue would like to skip creating github releases for their prereleases.

Could you rebase this PR on top of the current main branch? In the process please remove the `dist/index.js` file - since it's no longer committed to the repo.

src/index.ts
```diff
@@ -8,6 +8,7 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
 
 (async () => {
   let githubToken = process.env.GITHUB_TOKEN;
+  let skipCreateGithubRelease = Boolean(process.env.SKIP_CREATE_GITHUB_RELEASE);
```
Andarist, line 11: i think that the input option for the action would make more sense and I would call it `createGithubReleases`. It's default value would be `true` so you could disable it with `false`.

src/run.ts
```diff
@@ -106,10 +113,14 @@ export async function runPublish({
...
-          pkg,
-          tagName: `${pkg.packageJson.name}@${pkg.packageJson.version}`,
-        })
+        createRelease(
+          octokit,
+          {
+            pkg,
+            tagName: `${pkg.packageJson.name}@${pkg.packageJson.version}`,
+          },
+          skipCreateGithubRelease
```
Andarist, line 122: It imho makes more sense to call `createRelease` conditionally, rather than to skip releasing within this function.

akphi: The whole content of the `createRelease` function is about creating the Github release, so I used a `guarded if`, but perhaps, putting the whole thing in an `if` block is a bit more future proof 😅

src/index.ts
akphi: That's a fair point, I also figure this is a bit specific for this use case so polluting function param list with this boolean flag is not great, hence I wrapped it in `options`. If you think that's overkill, let me know.

akphi:
What about documentation, I felt like we should mention it somewhere if possible.

akphi:
Due to the change in base branch `master -> main`, I have to recreate the PR - https://github.com/changesets/action/pull/130

Andarist:
> Due to the change in base branch master -> main, I have to recreate the PR - #130

Just a tip - the target branch can usually be changed after clicking Edit. You can change the PR's title there but also the target branch.

PR #110, opened 2021-09-11
Add `authToken` for the npm registry to user-level `.npmrc` if it is not there

Andarist:
`.npmrc` can exist for unrelated reasons - for example `pnpm config set store-dir $PNPM_CACHE_FOLDER` could have been executed. In such a case we shouldn't bail out from creating the token for the npm registry because the action might fail unexpectedly (from the user's PoV).

From what I can tell the intention is to put this in the user-level `.npmrc` since a local file might exist and we want value from it to win. This doesn't change that anyhow and I think it's safe-to-do since the token was already auto-created in that file if the file itself was not existing. So any potential drawbacks of this approach would have already manifest themselves

I have thought about just prepending this to the user's `.npmrc` and it would be fine when it comes to loading values of this file and so on as specifying duplicate keys is not a problem. However, I've found a test case in the npm (linked in the code comment) which states that this can be added to the `.npmrc` under a few, slightly different, keys. So I had to go for the regexp-based detection solution.


Andarist:
@mitchellhamilton friendly 🏓

mitchellhamilton (approved):

PR #115, opened 2021-11-11
fix: parsing of .npmrc file

### LINKED ISSUE #114: Potential regression caused by #110
It seems that the latest changes merged from #110 are causing the release process to fail.

image

If I pin the GH action to the previous commit `b98cec97583b917ff1dc6179dd4d230d3e439894` instead of `changesets/action@master` it works fine.

Not sure how the changes related to `.npmrc` are causing this issue 🤔. @Andarist any idea maybe?

Also, I noticed that there is something odd with the if-else logic. In my setup I do have the `.npmrc` file with the auth token but I see the message log `Didn't find existing auth token for the npm registry in the user .npmrc file, creating one`.

Screen Shot 2021-11-11 at 09 52 14

Let me know if I can help solving this. Thanks

emmenko:
Fixes #114 

I decided to use the `npm/ini` package to take care of the parse/stringify part of the `.npmrc` config (which is what NPM itself uses). This avoids us having to deal with whitespaces etc.

To make it easier to test, I extracted the logic into a function in a separate file `npmUtils`.


src/npmUtils.ts
```diff
@@ -0,0 +1,50 @@
...
+  // Allow to inject a custom object, useful in tests.
+  processEnv = process.env
+) => {
+  const userNpmrcPath = `${processEnv.HOME}/.npmrc`;
+
+  if (fs.existsSync(userNpmrcPath)) {
+    console.log(`Found existing user .npmrc file at ${userNpmrcPath}`);
+
+    // Parse the `.npmrc` content using the `npm/ini` package.
+    const npmConfig = ini.parse(fs.readFileSync(userNpmrcPath, "utf-8"));
```
emmenko, line 14: We use `ini.parse` to take care of parsing the config (all whitespaces are stripped out)

src/npmUtils.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    let hasAuthToken = false;
+    for (const [key, value] of Object.entries(npmConfig)) {
+      if (/\/\/(.*)authToken$/.test(key) && Boolean(value)) {
+        console.log("The .npmrc file has an authToken");
+        hasAuthToken = true;
+      }
+    }
```
emmenko, line 22: This is the "new logic" to test the presence of the `authToken`. The regex also can be simpler.

src/npmUtils.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    if (!hasAuthToken) {
+      console.log(
+        "The .npmrc file does not have an authToken defined, creating one using the `NPM_TOKEN` environment variable"
+      );
+      if (!processEnv.NPM_TOKEN) {
```
emmenko, line 28: I thought it would make sense to throw an error if the `NPM_TOKEN` is not present.

src/npmUtils.ts
```diff
@@ -0,0 +1,50 @@
...
+        throw new Error(
+          "Missing NPM authToken. Please make sure you have the `NPM_TOKEN` environment variable defined."
+        );
+      }
+      npmConfig["//registry.npmjs.org/:_authToken"] = processEnv.NPM_TOKEN;
+      fs.writeFileSync(userNpmrcPath, ini.stringify(npmConfig));
```
emmenko, line 34: We can assign new values to the config object and then let `npm/ini` serialize the content.

Andarist:
Thank you for the PR! This refactor might be worth it but it will require some further analysis - if you are fine with waiting then we can leave it as is and I will get back to it this week. I would prefer to merge a small PR, that would only fix the whitespace issue, first. That could hit the stable release sooner.

emmenko:
Sure, I opened a separate PR with only a quick fix for the regex: #116

src/npmUtils.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+  } else {
+    console.log("No user .npmrc file found, creating one");
+    if (!processEnv.NPM_TOKEN) {
+      throw new Error(
+        "Missing NPM authToken. Please make sure you have the `NPM_TOKEN` environment variable defined."
+      );
+    }
```
Andarist, line 42: This should be removed. There is also a possibility to have `.npmrc` file in the project itself - it takes precedence over the user-level `.npmrc` and this should be respected. The current logic is that we try to add token to the user-level `.npmrc` which should be harmless if there is a project-level `.npmrc` because it will "shadow" over the user-level `.npmrc` anyway.

src/npmUtils.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  if (fs.existsSync(userNpmrcPath)) {
+    console.log(`Found existing user .npmrc file at ${userNpmrcPath}`);
+
+    // Parse the `.npmrc` content using the `npm/ini` package.
+    const npmConfig = ini.parse(fs.readFileSync(userNpmrcPath, "utf-8"));
+
+    let hasAuthToken = false;
+    for (const [key, value] of Object.entries(npmConfig)) {
+      if (/\/\/(.*)authToken$/.test(key) && Boolean(value)) {
```
Andarist, line 18: We should tighten up this regex - one, in theory, can have defined multiple auth tokens, for multiple registries. If one has a token for registry foo then we should still create an entry for the npm registry (if it's missing)

emmenko: Ah good point!

src/npmUtils.ts
emmenko: Hmm ok but then we end up writing an "empty/undefined" `authToken` value. Is that what we want?

Andarist: No, probably not - we can skip over creating/editing `.npmrc` when `NPM_TOKEN` is not set

README.md
```diff
@@ -10,6 +10,7 @@ This action for [Changesets](https://github.com/atlassian/changesets) creates a
 - version - The command to update version, edit CHANGELOG, read and delete changesets. Default to `changeset version` if not provided
 - commit - The commit message to use. Default to `Version Packages`
 - title - The pull request title. Default to `Version Packages`
+- skipNpmrcCheck - Check if there is a user-defined `.npmrc` file with a proper npm registry and an `authToken`. If not, a default `.npmrc` file is created.
```
emmenko, line 13: @Andarist I added this option to opt-out of the .npmrc checks.

I find it useful in case I don't need the use the default npm registry (for example when using github packages).

What do you think?

src/npmUtils.ts
```diff
@@ -0,0 +1,59 @@
+import fs from "fs";
+import * as ini from "ini";
+
+// https://docs.npmjs.com/using-private-packages-in-a-ci-cd-workflow#create-and-check-in-a-project-specific-npmrc-file
+const npmRegistryTokenKey = "//registry.npmjs.org/:_authToken";
```
emmenko, line 5: I did some research and couldn't find a case where the value isn't exactly like this (for the npm registry). Therefore I think we can omit the regex and simply do a 1:1 check.

src/npmUtils.ts
```diff
@@ -0,0 +1,59 @@
...
+    } else {
+      console.log(
+        "The .npmrc file does not have an authToken defined, appending one using the `NPM_TOKEN` environment variable..."
+      );
+      if (processEnv.NPM_TOKEN) {
+        npmConfig["//registry.npmjs.org/:_authToken"] = processEnv.NPM_TOKEN;
+        fs.writeFileSync(userNpmrcPath, ini.stringify(npmConfig));
+      } else {
+        console.warn(
+          "Missing `NPM_TOKEN` environment variable, skipping update of .npmrc file."
```
emmenko, line 40: So now in case the `NPM_TOKEN` is missing, we simply log a warning.

emmenko:
@Andarist Hey, I updated the PR based on the feedback. I also added a new option `skipNpmrcCheck` (see comment).

Let me know how you prefer to proceed here, thanks!

bluwy:
I'm going to close this for now as this merges to master while we're using main. But we're also looking into fixing this altogether soon in the next major.

PR #117, opened 2021-11-21
fix: correct logic of `switchToMaybeExistingBranch`

JounQin:
`stderr` will never contain `Switched to a new branch '${branch}'`, actually.


Andarist:
> stderr will never contain Switched to a new branch '${branch}', actually.

so this PR just removes the `' a '` from the expected string? I wonder if we couldn't use exit codes here to decouple ourselves completely from the text of this message

JounQin:
This PR removes `a new ` from the expected string because `git checkout` will never create a new branch. An exist code is similar with specific error message maybe, because an error message is more easier to read.

Andarist:
After thinking about this - it would be better/simpler to just work on the detached head and then force push that to a specific branch. Kinda like I'm doing it here: https://github.com/changesets/action/pull/118/files#diff-df56272098680154868a3f6139c97765b1599ba8d2a3f2f1ad72ce5be08163b6R30-R42

PR #118, opened 2021-11-22
Switch to Changesets-based release process

Andarist:
The goal of this PR is to have a good release process in place. After this lands we should:
- have a clear way to pin the action to a specific version
- should no longer have guaranteed conflicts when parallel work is happening on branches (cause `dist` wont be committed now)
- should increase the security (we were not reviewing `dist` before merging and a bad actor could tamper with this)


.github/workflows/version-or-publish.yml
```diff
@@ -0,0 +1,30 @@
+name: Version or Publish
+
+on:
+  push:
+    branches:
+      - main
```
Andarist, line 6: we should be able to rename the base branch now - we need to preserve `master` as artifact of the past though as all existing users are using that ref at the moment

scripts/release.js
```diff
@@ -0,0 +1,35 @@
+const path = require("path");
```
Andarist, line 1: I've done my best to test this before creating the PR - but as with any CI work... i could miss something and we'll have to supervise the first automated run of this script

scripts/release.js
```diff
@@ -0,0 +1,35 @@
...
+const { exec, getExecOutput } = require("@actions/exec");
+const tag = `v${require("../package.json").version}`;
```
Andarist, line 3: There is also a question - what version should we release first? I probably would say that we should immediately bump to v1 - leaving a possibility to add all missing tags retroactively in the future

.github/workflows/version-or-publish.yml
```diff
@@ -0,0 +1,30 @@
...
+
+jobs:
+  changesets:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@master
+
+      - name: Setup Node.js 12.x
```
emmenko, line 15: Shouldn't we use at least node 14?

.github/workflows/version-or-publish.yml
```diff
@@ -0,0 +1,30 @@
...
+        uses: actions/setup-node@master
+        with:
+          node-version: 12.x
+
+      - name: Install Dependencies
+        run: yarn
```
emmenko, line 21:
```suggestion
        run: yarn --frozen-lockfile
```

.github/workflows/version-or-publish.yml
```diff
@@ -0,0 +1,30 @@
...
+
```
emmenko, line 22: Add a step to build the code here?

.github/workflows/version-or-publish.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Create Release Pull Request or Publish
+        id: changesets
+        uses: changesets/action@master
```
emmenko, line 25: Shouldn't we point to the local built action?

```suggestion
        uses: .
```

This requires of course to build the code first.

.github/workflows/version-or-publish.yml
```diff
@@ -0,0 +1,30 @@
...
+        with:
+          version: yarn version
+          publish: yarn release
```
emmenko, line 28: Can you explain why do we need those two custom built scripts? 🤔

emmenko (commented):
Great idea, thanks for looking into this!

mitchellhamilton:
Could we add a `v${major}` tag? (and maybe `v${major}.${minor}` but I don't care very much about that) I don't really see the value in setting `changesets/action@v${major}.${minor}.${patch}` since it's bad for both staying up to date and ensuring the code you're using can't be tampered with, if you want to ensure that the code can't be tampered with, you should be using a commit and if you want to stay up to date, you should use a `v${major}` or `v${major}.${minor}` tag.

.github/workflows/version-or-publish.yml
Andarist: I'm not sure if that has changed since then but apparently noode 14 couldn't be used a few months back:
https://github.com/actions/github-script/pull/182#issuecomment-903966153

.github/workflows/version-or-publish.yml
Andarist: Probably - yes. I wasn't sure how to make it work though, are you sure that this will work?

.github/workflows/version-or-publish.yml
Andarist: build is happening within the release script - and it's only done **after** we check if we need to release at all, so this is a minor optimization of the execution run, since most of the runs wont need to release and thus wont have to run build

.github/workflows/version-or-publish.yml
Andarist: version:
- runs changeset version
- and updates README

release:
- checks if the current version has already been released
- builds
- commits
- tags the commit
- pushes the tag out

It's not a standard npm-based flow that is somewhat assumed by Changesets.

Andarist:
@mitchellhamilton Ok, makes sense - I've implemented the suggested change

scripts/release.js
```diff
@@ -0,0 +1,43 @@
...
+
+  await exec("changeset", ["tag"]);
+
+  await exec("git", [
+    "push",
+    "--force",
+    "--follow-tags",
+    "origin",
+    `HEAD:refs/heads/${releaseLine}`,
+  ]);
```
Andarist, line 42: note that i've chosen to use a branch for this because branches sync better than tags between remote and a local copy, to push out a tag with the same label we need to first destroy it and recreate. In general, I prefer to treat tags as immutable pointers.

mitchellhamilton:
The code looks fine to me, I'm thinking we should merge this to a new `main` branch, do a release and when we've tested it, change the default branch to `main` so we can make sure it works without breaking existing users if the new thing is initially broken? After that, we might want to add to the action on `master` to make it log a warning (as in https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#setting-a-warning-message) telling people to use the updated version.

Andarist:
@mitchellhamilton do u want to do it or should i go with it when i wake up?

mitchellhamilton:
Happy for you to do it

mitchellhamilton (approved):

.github/workflows/version-or-publish.yml
emmenko: Where is actions/github-script being used?

.github/workflows/version-or-publish.yml
emmenko: Yes it should work (see example here https://docs.github.com/en/actions/creating-actions/creating-a-javascript-action#example-using-a-private-action)

.github/workflows/version-or-publish.yml
Andarist: ah, i might have confused both things in a hurry - it might only not be possible to use node 14 for action itself but the setup-node action can actually setup node 14

rhmhsydjabr0-ctrl (commented):
#523 ovfezssh@gmail.com#
#523 Osama Atef#

krsjenswbp (commented):
thanks

PR #120, opened 2021-11-25
Deprecate `master` branch

Andarist:
⚠️ should only be marged after we switch to `main` as the default branch


src/index.ts
```diff
@@ -4,6 +4,13 @@ import * as gitUtils from "./gitUtils";
 import { runPublish, runVersion } from "./run";
 import readChangesetState from "./readChangesetState";
 
+core.warning(
```
Andarist, line 7: I've considered using `core.notice` - but I guess this is a "real" deprecation and that deserves a warning

src/index.ts
```diff
@@ -4,6 +4,13 @@ import * as gitUtils from "./gitUtils";
...
+  "The workflow file using `changesets/action` is currently using `@master` as the version. This branch has been frozen and deprecated. Please update your workflow to either use `@v1` or a specific commit SHA that is tagged.",
+  {
+    file: ".changeset/config.json",
```
Andarist, line 10: Without this, the warning can only be seen in the job's log or on the summary page of the whole workflow. This is not often visited though.

Annotating a specific file allows GitHub to display such annotation in the "Files changed" tab, below changes files. What is nice is that the annotated file doesn't have to be a part of the PR - the annotation might still be displayed in the "Unchanged files with check annotations" section. This is labeled with Beta - I'm not sure if it's enabled for all repositories etc. 

For this to be displayed successfully - the annotated file has to exist in the repo (which makes sense) and the only file that we know that should exist (and that makes sense to be annotated) is the `.changeset/config.json`

mitchellhamilton: The action isn't run on a PR though so wouldn't you still have to navigate to the action log to see it regardless of the changed files thing?

Andarist: Let's face it - I'm sometimes dumb.

Andarist: So basically this `core.warning` will be barely noticeable - still better than nothing 🤷‍♂️

mitchellhamilton (approved):

src/index.ts
Andarist: i've dropped this file annotation as it's pretty much useless

PR #125, opened 2021-12-16
make the branch name configurable

jacklesliewise:
a new optional input to change the branch name for the action PR

### DIFF
--- a/.changeset/eleven-pets-itch.md
+++ b/.changeset/eleven-pets-itch.md
@@ -1,5 +1,5 @@
 ---
-"@changesets/action": patch
+"@changesets/action": minor
 ---
 
 Make the branch name configurable with the optional `branch` input option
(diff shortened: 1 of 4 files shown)


Andarist:
What's the motivation for this feature? Why does it matter?

jacklesliewise:
Actually since opening this I've realised our use-case isn't actually required anymore 😅 

We had a github action job failing because of a CLI tool we were using was trying to create a folder based on the branch name `changeset-release/main` but since the `changeset-release` folder didn't exist it was failing.

Generally we avoid this by having branch names without `/` in them

However, the job we're running doesn't actually need to be run on this branch, so we can just not run the job on this branch

Andarist:
I guess we can consider using a different delimiter in the future to avoid situations like this. However, I would still classify this as a problem of that other job - since this is a valid branch name so it should... deal with it 😅

PR #127, opened 2021-12-18
add git commit signing

### LINKED ISSUE #126: Add option to enable commit signing
# Context

At the moment, if you use this action and have the `Require signed commits` setting enabled for a protected branch, the PR generated can't be merged.

# Example

[Crowdin](https://crowdin.com/), a popular localisation management tool, has a similar [action](https://github.com/crowdin/github-action) which opens a PR using a bot.

They allow you to pass a `gpg_private_key` as a config option, which will enable signed commits for the bot opening the PR. See options [here](https://github.com/crowdin/github-action#supported-options).

jacklesliewise:
# Context

Fixes #126 

# Changes

- Add a `githubUserName` input option
  - Default to `"github-actions[bot]"`
- Add a `githubUserEmail` input option
  - Default to `"github-actions[bot]@users.noreply.github.com"`
- Add a `GPG_PRIVATE_KEY` env option
  - This determines if keys should be signed or not
- Pass in `githubUserName` and `githubUserEmail` to `setupUser`
  - Can be done outwith signed commits, but these are needed to verify the signed commits if its enabled
- Add a `setupCommitSigning` if `GPG_PRIVATE_KEY` is present
  - Need to install `gpg`
  - use `STDIN` for `gpg --import` (utf-8 encoding with `Buffer.from` by default)
  - Need to do a `/bin/bash` command to get and grep the key id - `exec` doesn't support piping and multiple commands on its own
  - Set `git` config for `user.signingkey` and `commit.gpgsign`

# Considerations

- Input option naming (`githubUserName` vs `github_user_name`)
- Should the input options just be more `env` variables instead?]
- How do you feel about installing `gpg` here?

### DIFF
--- a/src/gitUtils.ts
+++ b/src/gitUtils.ts
@@ -1,20 +1,42 @@
 import { exec } from "@actions/exec";
 import { execWithOutput } from "./utils";
 
-export const setupUser = async () => {
+export const setupUser = async (githubUserName: string, githubUserEmail: string) => {
   await exec("git", [
     "config",
     "--global",
     "user.name",
-    `"github-actions[bot]"`,
+    githubUserName,
   ]);
   await exec("git", [
     "config",
     "--global",
     "user.email",
-    `"github-actions[bot]@users.noreply.github.com"`,
+    githubUserEmail,
   ]);
 };
+
+export const setupCommitSigning = async (gpgPrivateKey: string) => {
+  await exec("apk", [
+    "add",
+    "--no-cache",
+    "gnupg",
+  ]);
+  await exec("gpg", ["--import"], { input: Buffer.from(gpgPrivateKey) });
+  const { stdout: keyId } = await execWithOutput(`/bin/bash -c "gpg --list-secret-keys --with-colons | grep '^sec:' | cut -d ':' -f 5"`);
+  await exec("git", [
+    "config",
+    "--global",
+    "user.signingkey",
+    keyId.trim(),
+  ]);
+  await exec("git", [
+    "config",
+    "--global",
+    "commit.gpgsign",
+    "true",
+  ]);
+}
 
 export const pullBranch = async (branch: string) => {
   await exec("git", ["pull", "origin", branch]);
--- a/src/gitUtils.ts
+++ b/src/gitUtils.ts
@@ -1,42 +1,20 @@
 import { exec } from "@actions/exec";
 import { execWithOutput } from "./utils";
 
-export const setupUser = async (githubUserName: string, githubUserEmail: string) => {
-  await exec("git", [
-    "config",
-    "--global",
-    "user.name",
-    githubUserName,
-  ]);
-  await exec("git", [
-    "config",
-    "--global",
-    "user.email",
-    githubUserEmail,
-  ]);
+export const setupUser = async({ name, email }: { name: string; email: string; }) => {
+  await exec("git", ["config", "--global", "user.name", name]);
+  await exec("git", ["config", "--global", "user.email", email]);
 };
 
 export const setupCommitSigning = async (gpgPrivateKey: string) => {
-  await exec("apk", [
-    "add",
-    "--no-cache",
-    "gnupg",
-  ]);
+  await exec("apk", ["add", "--no-cache", "gnupg"]);
   await exec("gpg", ["--import"], { input: Buffer.from(gpgPrivateKey) });
-  const { stdout: keyId } = await execWithOutput(`/bin/bash -c "gpg --list-secret-keys --with-colons | grep '^sec:' | cut -d ':' -f 5"`);
-  await exec("git", [
-    "config",
-    "--global",
-    "user.signingkey",
-    keyId.trim(),
-  ]);
-  await exec("git", [
-    "config",
-    "--global",
-    "commit.gpgsign",
-    "true",
-  ]);
-}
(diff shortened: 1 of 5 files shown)


src/gitUtils.ts
```diff
@@ -1,21 +1,43 @@
...
     "config",
     "--global",
     "user.email",
-    `"github-actions[bot]@users.noreply.github.com"`,
+    githubUserEmail,
   ]);
 };
 
+export const setupCommitSigning = async (gpgPrivateKey: string) => {
+  await exec("apk", [
```
Andarist, line 20: is apk even available by default on ubuntu? why the Crowdin action doesn't have to do this installation at all?

src/gitUtils.ts
```diff
@@ -1,21 +1,43 @@
...
+    "add",
+    "--no-cache",
+    "gnupg",
+  ]);
+  await exec("gpg", ["--import"], { input: Buffer.from(gpgPrivateKey) });
+  const { stdout: keyId } = await execWithOutput(`/bin/bash -c "gpg --list-secret-keys --with-colons | grep '^sec:' | cut -d ':' -f 5"`);
```
Andarist, line 26: couldnt this be decoded from the importing's command output? something like this should display the information about what is being imported
```
gpg --with-colons --import-options show-only --import
```

the problem with listing all keys is that potentially the returned list could potentially contain many entries and we could grab some unrelated `keyId` here

src/index.ts
```diff
@@ -14,8 +14,19 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
     return;
   }
 
+  let githubUserName = getOptionalInput('githubUserName') || `"github-actions[bot]"`;
+  let githubUserEmail = getOptionalInput('githubUserEmail') || `"github-actions[bot]@users.noreply.github.com"`;
```
Andarist, line 18: those are not related to github, so I think this would be more appropriate:
```suggestion
  let gitUserName = getOptionalInput('gitUserName') || `"github-actions[bot]"`;
  let gitUserEmail = getOptionalInput('gitUserEmail') || `"github-actions[bot]@users.noreply.github.com"`;
```

src/index.ts
```diff
@@ -14,8 +14,19 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
...
+
   console.log("setting git user");
-  await gitUtils.setupUser();
+ 
+  await gitUtils.setupUser(githubUserName, githubUserEmail);
```
Andarist, line 22: nit: we could just group those nicely, like this:
```suggestion
  await gitUtils.setupUser({ name: gitUserName, email: gitUserEmail });
```

src/gitUtils.ts
jacklesliewise: The crowdin action does this installation in the [`Dockerfile`](https://github.com/crowdin/github-action/blob/master/Dockerfile)

What would the alternative be? Have a prerequisite that if you want commit signing the image you're using should have it installed already?

src/gitUtils.ts
jacklesliewise: Yeah true, although the current implementation seems to be sufficient for the crowdin action, especially if consumers are just passing the one private key.

Happy to explore your solution though if this is a blocker.

src/index.ts
jacklesliewise: Makes sense, will update these along with the README.

src/gitUtils.ts
Andarist: > What would the alternative be? Have a prerequisite that if you want commit signing the image you're using should have it installed already?

Yes, this sounds reasonable to me - I agree that it would be better if we could handle this automatically, but installing tools seems out of the scope of this simple node.js action. Note that most of our users use the default ubuntu images but it's technically possible that this action gets executed in different contexts too - so I'm afraid that we would always fail to recognize a particular environment and thus we'd fail to install this correctly on the user's behalf.

src/gitUtils.ts
Andarist: It's not a blocker per se but if this solution could be used then I would prefer using it - cause it seems to be more explicit about the whole flow. It's less dependent on other factors and the preexisting environment.

src/gitUtils.ts
jacklesliewise: Agreed, I'll remove this line then and add something to the `README`

src/gitUtils.ts
jacklesliewise: Yeah that makes sense, I won't pick this up for a few days now but I've addressed all the other comments, if it's sufficient as is then feel free to merge but if not I'll come back to this

mitchellhamilton:
Could we just have an option to disable setting the git email and name and then you can setup the signing before the Changesets action runs? It seems strange to have this built-in to the Changesets action.

jacklesliewise:
@mitchellhamilton yeah that's valid, I'm happy to open up a new PR with that approach if you prefer.

The current implementation is based on the Crowdin action, so it's up to you whether you'd like to align with that or if you'd rather have the action assume the git user is set up.

jacklesliewise:
@mitchellhamilton here's another [PR](https://github.com/changesets/action/pull/131) with the solution you suggested

Let me know which approach you both prefer

jacklesliewise:
Closing in favour of #131 131

PR #130, opened 2021-12-26
allow skipping creation of Github releases

akphi:
Reopen https://github.com/changesets/action/pull/108 due to change in base branch to `main`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,7 +8,6 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
 
 (async () => {
   let githubToken = process.env.GITHUB_TOKEN;
-  let createGithubReleases = process.env.CREATE_GITHUB_RELEASE;
 
   if (!githubToken) {
     core.setFailed("Please add the GITHUB_TOKEN to the changesets action");
@@ -76,10 +75,9 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
         script: publishScript,
         githubToken,
         options: {
-          createGithubReleases:
-            createGithubReleases !== undefined
-              ? Boolean(createGithubReleases)
-              : true,
+          createGithubReleases: JSON.parse(
+            getOptionalInput("createGithubReleases") ?? "true"
+          ),
         },
       });
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -75,9 +75,9 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
         script: publishScript,
         githubToken,
         options: {
-          createGithubReleases: JSON.parse(
-            getOptionalInput("createGithubReleases") ?? "true"
-          ),
+          createGithubReleases: getOptionalInput("createGithubReleases")
+            ? core.getBooleanInput("createGithubReleases")
+            : true,
         },
       });
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -74,11 +74,7 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
       const result = await runPublish({
         script: publishScript,
         githubToken,
-        options: {
-          createGithubReleases: getOptionalInput("createGithubReleases")
-            ? core.getBooleanInput("createGithubReleases")
-            : true,
-        },
+        createGithubReleases: core.getBooleanInput("createGithubReleases"),
       });
 
       if (result.published) {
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -74,6 +75,12 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
       const result = await runPublish({
         script: publishScript,
         githubToken,
+        options: {
+          createGithubReleases:
+            createGithubReleases !== undefined
+              ? Boolean(createGithubReleases)
+              : true,
+        },
```
Andarist, line 83: I would refactor this to:
```suggestion
        createGithubReleases: JSON.parse(getOptionalInput('createGithubReleases') ?? 'true'),
```

Any particular reason why you have implemented this as an env variable (`process.env.CREATE_GITHUB_RELEASE`) and not as an input of the action?

src/run.ts
```diff
@@ -17,9 +17,19 @@ import resolveFrom from "resolve-from";
...
+    createGithubReleases?: boolean;
+  } = {}
 ) => {
   try {
+    if (
+      options.createGithubReleases !== undefined &&
+      !options.createGithubReleases
+    ) {
+      return;
+    }
```
Andarist, line 31: I still kinda think that this should be moved outside of the `createRelease` function - if I call such a function I expect it to actually do what it is named after. Kinda a single-responsibility approach.

src/index.ts
akphi: Not really, I just didn't learn about this way 😓 

Fixed, but I start to wonder where we can document all these action inputs?

src/run.ts
akphi: Updated!

src/index.ts
Andarist: It should be added and documented~ in this file: https://github.com/changesets/action/blob/f4afbea27e0f68b63e3e018d053582abee1817b6/action.yml

The additional documentation about it in the README would also be cool but we can do that later.

src/index.ts
```diff
@@ -74,6 +74,11 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
       const result = await runPublish({
         script: publishScript,
         githubToken,
+        options: {
+          createGithubReleases: JSON.parse(
+            getOptionalInput("createGithubReleases") ?? "true"
+          ),
```
Andarist, line 80: I've just learned about this helper function that can probably be used here:
```suggestion
          createGithubReleases: core.getBooleanInput("createGithubReleases"),
```

I believe that it takes the default declared in the `.yml` file into consideration but it would be great if we could confirm that.

akphi: I don't think this is safe to do. That function (right from its type-def returns a boolean), but this is optional.

https://github.com/actions/toolkit/blob/daf8bb00606d37ee2431d9b1596b88513dcf9c59/packages/core/src/core.ts#L171-L182

In case people don't provide this (which is pretty much for all current users), it would throw error.

src/index.ts
akphi: Added document in the `action.yml` file

src/index.ts
Andarist: I think that if it's not safe to do then we've made a horrible mistake merging this: https://github.com/changesets/action/pull/131/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R17 . While merging I didn't notice that it's using this `getBooleanInput` and I didn't check it back then. If this would throw then we'd already have some angry users in the issues about broken workflow runs :sweat_smile:

I think that the defaults might be provided by the workflow runner using environment variables:
https://github.com/actions/toolkit/blob/daf8bb00606d37ee2431d9b1596b88513dcf9c59/packages/core/src/core.ts#L129-L130

I'll try to recheck this later in my private playground repo.

akphi: Could also be because people are using `changesets/action@v1` like in the `README.md` file 😅
A bit unrelated, do you know why this action doesn't show up in Github action marketplace, i.e. why don't we have this in our repo page?

image

Andarist: > Could also be because people are using changesets/action@v1 like in the README.md file 

I've rechecked in my private github actions playground and it seems like a call to `core.getBooleanInput("createGithubReleases")` is enough :)

> A bit unrelated, do you know why this action doesn't show up in Github action marketplace, i.e. why don't we have this in our repo page?

Not sure, I'm not sure what's the requirement for this.

Andarist:
@akphi thanks for being patient and bearing with me!

PR #131, opened 2021-12-29
custom git user input option

### LINKED ISSUE #126: Add option to enable commit signing

jacklesliewise:
# Context

Fixes #126, alternative solution to #127

Instead of implementing git commit signing into the action, we could allow an input option to disable the `setupUser` function so that consumers can use their own git user to create commits in this action.

# Changes

- Add `setupGitUser` boolean input option
  - Updated `@actions/core` to be able to use `getBooleanInput`
  - Default to `true`
- Don't run `gitUtils.setupUser()` if `setupGitUser` is `false`
- Update docs

### DIFF
--- a/.changeset/two-dolls-promise.md
+++ b/.changeset/two-dolls-promise.md
@@ -2,4 +2,4 @@
 "@changesets/action": minor
 ---
 
-add a `setupGitUser` to enable or disable setting up a default git user
+Added `setupGitUser` option to enable or disable setting up a default git user
--- a/.changeset/two-dolls-promise.md
+++ b/.changeset/two-dolls-promise.md
@@ -1,0 +1,5 @@
+---
+"@changesets/action": patch
+---
+
+add an input option to allow custom git users
--- a/.changeset/two-dolls-promise.md
+++ b/.changeset/two-dolls-promise.md
@@ -2,4 +2,4 @@
 "@changesets/action": minor
 ---
 
-add an input option to allow custom git users
+add a `setupGitUser` to enable or disable setting up a default git user
--- a/.changeset/two-dolls-promise.md
+++ b/.changeset/two-dolls-promise.md
@@ -1,5 +1,5 @@
 ---
-"@changesets/action": patch
+"@changesets/action": minor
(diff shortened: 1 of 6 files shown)


jacklesliewise:
@Andarist @mitchellhamilton just bumping this if that's okay

Would you prefer this approach or the one in #127?

action.yml
```diff
@@ -17,6 +17,9 @@ inputs:
   title:
     description: The pull request title. Default to `Version Packages`
     required: false
+  customGitUser:
```
Andarist, line 20: I would prefer for this to be called something like `setupGitUser` - with the default being `true`.

Andarist:
@jacklesliewise the only concern I have with this PR is the nit about the name that has been chosen for the option. I think this might be useful to some people in other situations than the one covered by #127 so this should land either way (regardless of what we decide for the #127 but most likely we'll close it, for the time being, we can always revisit this in the future).

jacklesliewise:
Thanks @Andarist, I've closed #127 for now and updated this PR as per your comment

action.yml
jacklesliewise: Sounds good, updated

.changeset/two-dolls-promise.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/action": minor
+---
+
+add a `setupGitUser` to enable or disable setting up a default git user
```
mitchellhamilton, line 5:
```suggestion
Added `setupGitUser` option to enable or disable setting up a default git user
```

mitchellhamilton (approved):

Issue #135, opened 2022-01-19
Feature Request: Expose hasChangesets


wweaver:
I'm looking for a way to create tags at the time of publish instead of actually publishing. I'm working with a monorepo but we don't publish to npm. We instead are trying to use this as a deploy job. The creation of a version PR works nicely but I'm trying to find a simple way to hook into when the version PR is merged to generate the corresponding tags but not perform the publish. I thought the simplest way to do this would be to provide the additional output from this script so I can add this new flow.

```yaml
            - name: Create Release Pull Request
              uses: changesets/action@v1
              env:
                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}

            - name: Create Tags
              if: "!steps.changesets.output.hasChangesets"
              run: pnpm changeset tag && git push --tags
```

If there's a better way I'm all for any input!

I had tried to do this as I thought this was overriding the publish command and I'd have to call `changeset publish` manually, but it seems like it's a separate script and the `changeset publish` is called automatically anyway.

```yaml
            - name: Create Release Pull Request
              uses: changesets/action@v1
              with:
                  publish: pnpm publish
              env:
                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

PR #136, opened 2022-01-19
Add shouldPublish output

wweaver:
This is untested but I believe if I understand the code correctly should work.

This will provide a `steps.changesets.output.shouldPublish` variable to use by subsequent commands in a github workflow

closes: #135


src/index.ts
```diff
@@ -30,18 +30,19 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
...
 
   let publishScript = core.getInput("publish");
-  let hasChangesets = changesets.length !== 0;
+  let shouldVersion = changesets.length !== 0;
+  let shouldPublish = changesets.length === 0;
   let hasPublishScript = !!publishScript;
 
   core.setOutput("published", "false");
   core.setOutput("publishedPackages", "[]");
-  core.setOutput("hasChangesets", String(hasChangesets));
```
Andarist, line 38: isnt this just changing the name? the functionality you are looking for is already here

wweaver: I took some liberties which I can understand reverting. The only change that really needs to be made is adding the variable to the actions.yml

Andarist: Ye, i would prefer it if u could bring back the previous name since renaming that would be a breaking change for me 😅 

https://github.com/statelyai/xstate-viz/blob/259fb02e170017139cc735f87e673e2125dcfb06/.github/workflows/version-or-release.yml#L39

wweaver: Interesting. So you're saying it works for you already? Wonder why it wasn't working for me.

I can just add the readme change and revert back to hasChangesets. Probably won't be until tomorrow. Thanks for the feedback!

Andarist: > So you're saying it works for you already?

I hope so 🤣 

> I can just add the readme change and revert back to hasChangesets. Probably won't be until tomorrow. Thanks for the feedback!

No worries, just leave edits to README and action.yml

wweaver: Updated

wweaver:
@Andarist shall I close this PR? If it's not what you'd want I'd rather just close it instead of leaving it open.

README.md
```diff
@@ -123,6 +123,50 @@ For example, you can add a step before running the Changesets GitHub Action:
     NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
 ```
 
+#### Custom Publishing
+
+If you want to hook into when publishing should occur but have your own publishing functionality you can 
+utilize the `hasChangesets` output.
```
Andarist, line 129:
```suggestion
If you want to hook into when publishing should occur but have your own publishing functionality you can utilize the `hasChangesets` output.

Not that you might need to account for things already being published in your script because a commit without any new changesets can always land on your base branch after a successful publish. In such a case you need to figure out on your own how to skip over the actual publishing logic or handle errors gracefully as most package registries won't allow you to publish over already published version.
```

.changeset/hip-eagles-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/action": patch
+---
+
+Provider steps.changesets.output.shouldPublish to github workflow
```
Andarist, line 5: In a way this is just a docs PR so let's not add a changeset here.
```suggestion
---
---
```

Andarist:
@wweaver sorry for not responding sooner - I just have a lot of notifications and sometimes stuff gets lost in all of that. I appreciate a friendly ping :)

.changeset/hip-eagles-sleep.md
wweaver: Removed. This was left over from when I was making more drastic changes. Thanks.

wweaver:
> @wweaver sorry for not responding sooner - I just have a lot of notifications and sometimes stuff gets lost in all of that. I appreciate a friendly ping :)

Thanks for looking. I made your suggested changes.

wweaver:
Thanks!

PR #142, opened 2022-01-24
fix output flag `published` is always `false` when used with `changeset publish --no-git-tag`

### LINKED ISSUE #141: Bug: output `published` is always `false` when used with `changeset publish --no-git-tag`
@Andarist While testing out my change in #130 I found out that when I set my publish script to use `changeset publish --no-git-tag`, the output `published` is always `false`. This is due to the following piece of code:

https://github.com/changesets/action/blob/e8ffc2d207b7b66f56931b03bc801c0c0a8dadce/src/run.ts#L145-L155

`releasedPackages` is only populated when the text `New tag:` is found in the CLI output. I'm not sure if this is still the right indicator for successful publishing, instead, I think we should fix it so that as long as no error thrown when running the publishing script, we should set the output `published` to `true`.

I believe a lot of people depends on this output flag in their github actions script, especially to distinguish between `publish flow` vs. `creating the new version PR` flow. I will try to create a new PR to address this, let me know what you think. Thanks!

akphi:
Fix #141

### DIFF
--- a/src/run.ts
+++ b/src/run.ts
@@ -57,14 +57,10 @@ type PublishOptions = {
 
 type PublishedPackage = { name: string; version: string };
 
-type PublishResult =
-  | {
-      published: true;
-      publishedPackages: PublishedPackage[];
-    }
-  | {
-      published: false;
-    };
+type PublishResult = {
+  published: boolean;
+  publishedPackages: PublishedPackage[];
+};
 
 export async function runPublish({
   script,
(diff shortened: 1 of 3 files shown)


src/run.ts
```diff
@@ -142,17 +138,16 @@ export async function runPublish({
     }
   }
 
+  const result: PublishResult = { published: true, publishedPackages: [] };
```
akphi, line 141: I thought a lot about this. This somewhat changes the meaning of `published`, with this, it means: as long as the publishing script runs fine, we consider `published=true`, in a way, this is coherent with the definition of the flag in our README file: `A boolean value to indicate whether a publishing is happened or not`. Like I said in the issue created, I think the main issue here is we need a flag to tell us in the Github action script that whether `publishing` or `versioning (creating the New Version PR)` happened.

Andarist: I agree that relying on the `New tag:` output is less than ideal. However, you can (sort of) check today which script has been executed by inspecting the `hasChangesets` output:
https://github.com/changesets/action/pull/136/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R165

akphi: hm, thinking about it, looks like my approach is quite wrong though. Will close this PR.

PR #154, opened 2022-03-01
Updated Node.js version to 16.x in README examples

trivikr:
Followed [Usage](https://github.com/actions/setup-node#usage) for https://github.com/actions/setup-node


Andarist:
I'm a little bit worried that this could be interpreted as if node 16 is a requirement for this action

trivikr:
Done in https://github.com/changesets/action/commit/595655c3eae7136ff5ba18200406898904362926

PR #157, opened 2022-03-04
fix: github PR body max length

emmenko:
I had an issue today where the PR body exceeded a size limit.

image

This prevents the PR to be created.

To fix that we can truncate the message if we detect it's too long.

Let me know what you think and if I need to add a test for this use case.

This is how the truncated message in the PR looks like:

image

### DIFF
--- a/src/run.test.ts
+++ b/src/run.test.ts
@@ -157,4 +157,114 @@ describe("version", () => {
 
     expect(mockedGithubMethods.pulls.create.mock.calls[0]).toMatchSnapshot();
   });
+
+  it("does not include changelog entry if message exceeds size limit", async () => {
+    let cwd = f.copy("simple-project");
+    linkNodeModules(cwd);
+
+    mockedGithubMethods.search.issuesAndPullRequests.mockImplementationOnce(
+      () => ({ data: { items: [] } })
+    );
+
+    await writeChangesets(
+      [
+        {
+          releases: [
+            {
+              name: "simple-project-pkg-a",
+              type: "minor",
+            },
+          ],
+          summary: `# Non manus superum
+
+## Nec cornibus aequa numinis multo onerosior adde
+
+Lorem markdownum undas consumpserat malas, nec est lupus; memorant gentisque ab
+limine auctore. Eatque et promptu deficit, quam videtur aequa est **faciat**,
+locus. Potentia deus habebat pia quam qui coniuge frater, tibi habent fertque
+viribus. E et cognoscere arcus, lacus aut sic pro crimina fuit tum **auxilium**
+dictis, qua, in.
+
+In modo. Nomen illa membra.
+
+> Corpora gratissima parens montibus tum coeperat qua remulus caelum Helenamque?
+> Non poenae modulatur Amathunta in concita superi, procerum pariter rapto cornu
+> munera. Perrhaebum parvo manus contingere, morari, spes per totiens ut
+> dividite proculcat facit, visa.
+
+Adspicit sequitur diffamatamque superi Phoebo qua quin lammina utque: per? Exit
+decus aut hac inpia, seducta mirantia extremo. Vidi pedes vetus. Saturnius
+fluminis divesque vulnere aquis parce lapsis rabie si visa fulmineis.
+`,
+        },
+      ],
+      cwd
+    );
+
+    await runVersion({
+      githubToken: "@@GITHUB_TOKEN",
+      cwd,
+      maxCharactersPerMessage: 1000,
+    });
+
+    expect(mockedGithubMethods.pulls.create.mock.calls[0]).toMatchSnapshot();
+    expect(mockedGithubMethods.pulls.create.mock.calls[0][0].body).toMatch(
+      /The changelog information of each package has been omitted from this message/
+    );
+  });
+
+  it("does not include any release information if message exceeds size limit", async () => {
+    let cwd = f.copy("simple-project");
+    linkNodeModules(cwd);
+
+    mockedGithubMethods.search.issuesAndPullRequests.mockImplementationOnce(
+      () => ({ data: { items: [] } })
+    );
+
+    await writeChangesets(
+      [
+        {
+          releases: [
+            {
+              name: "simple-project-pkg-a",
+              type: "minor",
+            },
+          ],
+          summary: `# Non manus superum
+
+## Nec cornibus aequa numinis multo onerosior adde
+
+Lorem markdownum undas consumpserat malas, nec est lupus; memorant gentisque ab
+limine auctore. Eatque et promptu deficit, quam videtur aequa est **faciat**,
+locus. Potentia deus habebat pia quam qui coniuge frater, tibi habent fertque
+viribus. E et cognoscere arcus, lacus aut sic pro crimina fuit tum **auxilium**
+dictis, qua, in.
+
+In modo. Nomen illa membra.
+
+> Corpora gratissima parens montibus tum coeperat qua remulus caelum Helenamque?
+> Non poenae modulatur Amathunta in concita superi, procerum pariter rapto cornu
+> munera. Perrhaebum parvo manus contingere, morari, spes per totiens ut
+> dividite proculcat facit, visa.
+
+Adspicit sequitur diffamatamque superi Phoebo qua quin lammina utque: per? Exit
+decus aut hac inpia, seducta mirantia extremo. Vidi pedes vetus. Saturnius
+fluminis divesque vulnere aquis parce lapsis rabie si visa fulmineis.
+`,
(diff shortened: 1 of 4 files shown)


src/run.ts
```diff
@@ -278,22 +284,33 @@ ${
 
   let searchResult = await searchResultPromise;
   console.log(JSON.stringify(searchResult.data, null, 2));
+
+  let messageBody = await prBodyPromise;
```
Andarist, line 288: Since we have `prBodyPromise`, this variable probably should be called `prBody` to make the connection between them obvious.

src/run.ts
```diff
@@ -278,22 +284,33 @@ ${
...
+  if (messageBody.length > MAX_CHARACTERS_PER_MESSAGE) {
+    console.log(
+      `message body too long, truncating at ${MAX_CHARACTERS_PER_MESSAGE} charachters.`
+    );
+    messageBody = [
+      messageBody.substring(0, MAX_CHARACTERS_PER_MESSAGE),
+      "[Too long, message truncated]",
+    ].join("\n\n");
```
Andarist, line 296: This has the potential of breaking some formatting stuff or breaking links etc (which has probably happened on the attached screenshot). I wonder if on such an occasion, we shouldn't just replace the whole "changelogs" part with a placeholder message explaining the problem and advising to read the git diff, we could even construct a link to the "File Changed" tab there. WDYT?

emmenko: Yes, formatting might be broken. I don't have a strong opinion to be honest.
The important part is that the PR is opened without errors, as otherwise the release is blocked.

Up to you to decide.

Andarist: I think it's safer to start with the suggested way.

We'd have to just move this part to compute a separate object:
https://github.com/changesets/action/blob/7e613b338ee4b15ca9b286bbd31bd1dd31e75594/src/run.ts#L245

and construct the final body later with the algorithm like:
1. prepare the message without considering the limit
2. if the limits has been exceeded then construct a fallback body that would have those things in it:
- ["this prefece"](https://github.com/changesets/action/blob/7e613b338ee4b15ca9b286bbd31bd1dd31e75594/src/run.ts#L226)
- information that the limits has been exceeded
- list of releases (names + versions, similar to what we already print in the full body, just without the changelog part

src/run.ts
```diff
@@ -168,13 +175,83 @@ const requireChangesetsCliPkgJson = (cwd: string) => {
...
+    highestLevel: number;
+    private: boolean;
+    content: string;
+    header: string;
+  }[];
+  maxCharactersPerMessage: number;
+  preState?: PreState;
+};
+
+export async function getChangesetsMessage({
```
emmenko, line 191: Extracted the logic to construct the message into a separate function.

src/run.ts
emmenko: @Andarist I pushed some changes following your suggestions. It would be nice if you can take a look when you have time.
Thanks

src/run.ts
```diff
@@ -168,13 +175,83 @@ const requireChangesetsCliPkgJson = (cwd: string) => {
...
+      messagePrestate,
+      messageReleases,
+      `\n> The changelog information of each package has been omitted from this message, as the content exceeds the size limit.\n`,
+      ...changedPackagesInfo.map((info) => `${info.header}\n\n`),
+    ].join("\n");
+  }
+
+  // Check (again) that the message is within the size limit.
+  // If not, omit all release content this time.
+  if (fullMessage.length > maxCharactersPerMessage) {
```
Andarist, line 235: heh, this is probably very, very unlikely :P but I appreciate thinking about this edge case 👍

src/run.ts
```diff
@@ -168,13 +175,83 @@ const requireChangesetsCliPkgJson = (cwd: string) => {
...
+`;
+  let messagePrestate = !!preState
+    ? `⚠️⚠️⚠️⚠️⚠️⚠️
+
+\`${branch}\` is currently in **pre mode** so this branch has prereleases rather than normal releases. If you want to exit prereleases, run \`changeset pre exit\` on \`${branch}\`.
+
+⚠️⚠️⚠️⚠️⚠️⚠️
+`
+    : "";
+  let messageReleases = `# Releases`;
```
Andarist, line 212: nit: 
```suggestion
  let messageReleasesHeading = `# Releases`;
```

src/run.ts
```diff
@@ -168,13 +175,83 @@ const requireChangesetsCliPkgJson = (cwd: string) => {
...
+}
+
 type VersionOptions = {
   script?: string;
   githubToken: string;
   cwd?: string;
   prTitle?: string;
   commitMessage?: string;
   hasPublishScript?: boolean;
+  maxCharactersPerMessage?: number;
```
Andarist, line 254: nit:
```suggestion
  prBodyMaxCharacters?: number;
```

src/run.ts
Andarist, line 191: nit:
```suggestion
export async function getVersionPrBody({
```

src/run.ts
```diff
@@ -214,55 +292,29 @@ export async function runVersion({
     q: searchQuery,
   });
   let changedPackages = await getChangedPackages(cwd, versionsByDirectory);
+  let changedPackagesInfo = (
+    await Promise.all(
```
Andarist, line 296: nit: previously this was parallelized, we could maintain this with:
```suggestion
    Promise.all(
```

src/run.ts
```diff
@@ -278,22 +330,31 @@ ${
 
   let searchResult = await searchResultPromise;
   console.log(JSON.stringify(searchResult.data, null, 2));
+
+  let prBody = await getChangesetsMessage({
+    hasPublishScript,
+    preState,
+    branch,
+    changedPackagesInfo,
```
Andarist, line 338: If you follow up on the comment [here](https://github.com/changesets/action/pull/157/files#r851184758) then this change will be required too:
```suggestion
    changedPackagesInfo: await changedPackagesInfo,
```

emmenko:
@Andarist Hi 👋, I applied your suggestions and rebased the branch.

I think this should be good to go now, unless you have other suggestions. Thanks

emmenko:
Btw, here's a real example of how this looks when the content is reduced.

https://github.com/commercetools/ui-kit/pull/2178

emmenko:
@Andarist Friendly reminder in case you haven't had a chance to look at the final changes. Thanks

src/run.test.ts
```diff
@@ -157,4 +157,114 @@ describe("version", () => {
 
     expect(mockedGithubMethods.pulls.create.mock.calls[0]).toMatchSnapshot();
   });
+
+  it("does not include changelog entry if message exceeds size limit", async () => {
```
Andarist, line 161:
```suggestion
  it("does not include changelog entries if full message exceeds size limit", async () => {
```

src/run.test.ts
```diff
@@ -157,4 +157,114 @@ describe("version", () => {
...
+      maxCharactersPerMessage: 1000,
+    });
+
+    expect(mockedGithubMethods.pulls.create.mock.calls[0]).toMatchSnapshot();
+    expect(mockedGithubMethods.pulls.create.mock.calls[0][0].body).toMatch(
+      /The changelog information of each package has been omitted from this message/
+    );
+  });
+
+  it("does not include any release information if message exceeds size limit", async () => {
```
Andarist, line 216:
```suggestion
  it("does not include any release information if a message with simplified release info exceeds size limit", async () => {
```

src/run.ts
```diff
@@ -159,6 +166,7 @@ const requireChangesetsCliPkgJson = (cwd: string) => {
   try {
     return require(resolveFrom(cwd, "@changesets/cli/package.json"));
   } catch (err) {
+    // @ts-ignore
```
Andarist, line 169:
```suggestion
  } catch (err: any) {
```

emmenko:
@Andarist Awesome, thanks!

tedraykov:
This solution only fixes the generated Version PR body exceeding the max body length. The problem still exists for creating GitHub releases:
```
Error: Validation Failed: {"resource":"Release","code":"custom","field":"body","message":"body is too long (maximum is 125000 characters)"}
```

The failing action run: https://github.com/tedraykov/reaction/actions/runs/3196539005/jobs/5218616158#step:7:163

Andarist:
My educated guess here is that we could detect the format of [this CHANGELOG](https://github.com/tedraykov/reaction/blob/138583d5707b30b5de7472b7ea4caebcb515db76/apps/reaction/CHANGELOG.md) correctly and thus the whole thing was passed as the github release body in the payload. This is a problem on its own. Exceeding the character limit with legitimate requests is possible but very unlikely.

Issue #159, opened 2022-03-08
git user setup fails unless the workspace root directory is also a git repo.


bicknellr:
The git user setup is run before changing directories for `cwd`. This causes the action to fail if the root directory is not a git repo, even if the intended directory for the action is not the root.

https://github.com/changesets/action/blob/5766bae8d9257554fd1f3b5936dbc49d595b4854/src/index.ts#L17-L34

Andarist:
I'm not sure if I understand the issue in full. Could you explain this like I'm five? Do you know how to fix this on our end?

The git user setup doesn't touch `cwd` so that's why I'm confused about this issue.

bicknellr:
Sorry, I phrased this issue poorly. It's not that changing the directory itself fails, it's that the `git config user.name whatever` fails unless the root workspace directory is a git repo because the action tries to run that git command before changing directories.

For example, lets say you're writing a workflow that checks out your repo to some path other than the root:

```
jobs:
  job-name:
    steps:
      - uses: actions/checkout@v3
        with:
          repository: some/repo
          path: some/path

      - uses: changesets/action@v1
        with:
          cwd: some/path
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

After the checkout action, `$GITHUB_WORKSPACE/some/path` is a git repo but `$GITHUB_WORKSPACE` isn't. Then, when the changesets action runs, the first thing it tries to do (i.e. before changing directories to `cwd`) is `git config user.name "github-actions[bot]"`, which fails because it is running in `$GITHUB_WORKSPACE`.

I think the solution is just to move this:

https://github.com/changesets/action/blob/5766bae8d9257554fd1f3b5936dbc49d595b4854/src/index.ts#L30-L34

Before this:

https://github.com/changesets/action/blob/5766bae8d9257554fd1f3b5936dbc49d595b4854/src/index.ts#L17-L22

Andarist:
Ok, thanks for the explanation. This makes sense to me now. Would you like to introduce such a change to the codebase?

bicknellr:
Sure, PR here: https://github.com/changesets/action/pull/161

PR #161, opened 2022-03-14
Change directory to `cwd` before running git user setup.

bicknellr:
This PR applies `cwd` before setting the git user name and email. Without this change, the git setup commands are run in `$GITHUB_WORKSPACE`, which will fail if `$GITHUB_WORKSPACE` is not a git repo.

Fixes https://github.com/changesets/action/issues/159.

### DIFF
--- a/.changeset/old-socks-unite.md
+++ b/.changeset/old-socks-unite.md
@@ -2,4 +2,4 @@
 "@changesets/action": patch
 ---
 
-Change directory to `cwd` before running git user setup.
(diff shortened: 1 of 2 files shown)


.changeset/old-socks-unite.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/action": patch
```
bicknellr, line 2: Just want to point out I've marked this as 'patch'.

Andarist (approved):

Issue #82, opened 2021-03-26
Include pull request number in action output


jonrohan:
👋🏻 Hi I would like to have the `pull_number` in the `steps.changesets.outputs.pull_number` output from the action step if `steps.changesets.outputs.published == "false"`. 

This is so I can have my action also post a status on the pull request after the changesets/action step runs. 

https://github.com/changesets/action/blob/aacbc223926b9a6ac55d5922b90fd2a09e1fcfa9/src/run.ts#L274-L291

Andarist:
Seems reasonable - I'm open for accepting a PR for this.

PR #167, opened 2022-04-05
feat: set pullRequestNumber as action output

dmregister:
Closes: https://github.com/changesets/action/issues/82

### DIFF
--- a/.changeset/long-mayflies-yell.md
+++ b/.changeset/long-mayflies-yell.md
@@ -2,4 +2,4 @@
 "@changesets/action": minor
 ---
 
-Add `pullRequestNumber` to the output of the action
+Added `pullRequestNumber` to the action's outputs
--- a/.changeset/long-mayflies-yell.md
+++ b/.changeset/long-mayflies-yell.md
@@ -1,0 +1,5 @@
+---
+"@changesets/action": minor
+---
+
+Add `pullRequestNumber` to the output of the action
--- a/src/index.ts
+++ b/src/index.ts
@@ -97,13 +97,16 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
       return;
     }
     case hasChangesets:
-      await runVersion({
+      const { pullRequestNumber } = await runVersion({
         script: getOptionalInput("version"),
         githubToken,
         prTitle: getOptionalInput("title"),
         commitMessage: getOptionalInput("commit"),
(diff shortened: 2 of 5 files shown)


src/run.ts
```diff
@@ -177,14 +177,18 @@ type VersionOptions = {
   hasPublishScript?: boolean;
 };
 
+type RunVersionResponse = {
```
Andarist, line 180: nit: it sounds like a response from some HTTP-like call to me but it's just a return value of an async function:
```suggestion
type RunVersionResult = {
```

src/index.ts
```diff
@@ -97,13 +97,16 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
...
-      await runVersion({
+      const { pullRequestNumber } = await runVersion({
         script: getOptionalInput("version"),
         githubToken,
         prTitle: getOptionalInput("title"),
         commitMessage: getOptionalInput("commit"),
         hasPublishScript,
       });
+
+      core.setOutput("pullRequestNumber", String(pullRequestNumber));
```
Andarist, line 108: Since this will always be int - could we not cast it with `String()`? Would it have any observable difference if this wouldn't be casted? Like perhaps one version would not be quoted or smth? Did you test this from the consumer's side of this action?

I've checked the source code of `@actions/core` and I think this gets stringified anyway but it would be great if we could confirm if there is any practical difference between both flavours - if there is not then I'm ok with leaving it as is.

src/run.ts
dmregister: done

.changeset/long-mayflies-yell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/action": minor
+---
+
+Add `pullRequestNumber` to the output of the action
```
Andarist, line 5:
```suggestion
Added `pullRequestNumber` to the action's outputs
```

Andarist: I've wanted to commit this suggestion through the GitHub interface - but it seems like I don't have sufficient permissions to push to your fork.

You can check how to allow this for the future PRs here - https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork . At the moment just commit this suggestion with a click on a button.

PR #173, opened 2022-04-14
Add Optional skipNpmrc input

JP-Dhabolt:
As part of trying to use this action for a private Azure DevOps Artifact Feed (where the `.npmrc` file uses `username` and `_password` instead of `_authToken`), I ran into the issue where the file was always modified by the action.

This PR adds an optional input called `skipNpmrc` which simply bypasses any checks, creation, or update of the `.npmrc` file.  Not including the new input continues to behave exactly as it does today.

None of the code in the new `createOrUpdateNpmrc` function was modified, just extracted into the function.

While I am unblocked since I am running the action successfully now from the fork, I thought this might be generally useful if others have the same problem.  If this is not in line with your vision of the action, please feel free to close.  No hard feelings.  😄

### DIFF
--- a/.changeset/tricky-impalas-protect.md
+++ b/.changeset/tricky-impalas-protect.md
@@ -1,0 +1,6 @@
+---
+"@changesets/action": minor
+---
+
+Added option to ignore creation or modification of `.npmrc` file.
+This resolves issues where the correctly configured `.npmrc` file does not match the assumptions made by the action, such as when authenticating to Azure DevOps Artifact Feeds.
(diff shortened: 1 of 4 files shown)


jameslnewell:
This would benefit our use case with AWS CodeArtifact too.

We have a `.npmrc` file that contains
```
@scope:registry=https://abcd.d.codeartifact.ap-southeast-2.amazonaws.com/npm/npm/
//abcd.d.codeartifact.ap-southeast-2.amazonaws.com/npm/npm/:_authToken=XXXX
```

And `changesets` is overwriting `.npmrc` to contain
```
//registry.npmjs.org/:_authToken=undefined
```

JP-Dhabolt:
@jameslnewell  If you update your workflow that references this action to `uses: JP-Dhabolt/changesets-actions@skip-npmrc-compiled`, it should act as a workaround until this gets merged in.

Issue #97, opened 2021-07-04
Documentation suggesting `fetch-depth: 0` is out of date


RoystonS:
The documentation for the action currently suggests this for the configuration:

>  # This makes Actions fetch all Git history so that Changesets can generate changelogs with the correct commits
>           fetch-depth: 0
> 

This is no longer necessary: the tool will automatically deepen shallow clones as necessary since https://github.com/atlassian/changesets/pull/495.

Andarist:
That is true - would you like to prepare a documentation change for this?

PR #177, opened 2022-04-28
Remove suggestions of using `fetch-depth: 0`

RoystonS:
Changesets automatically deepens a shallow clone as necessary,
so we don't need to recommend a complete deep clone in order
for changesets to work.

Fixes: #97


RoystonS:
No changeset is needed for this PR: it doesn't include any change which should result in a package release. It only changes documentation visible in this GitHub repo.

PR #180, opened 2022-05-07
chore(docs): Update README to reference baseBranch, use "main" as default in examples

brianespinosa:
- Adds reference to the `baseBranch` config in the Changesets docs
- Updates all references in examples to `main` instead of `master`
  - [`main` is now the default branch](https://github.com/github/renaming) in new GitHub projects
  - This change aligns with the comment in the `baseBranch` section of Changesets docs


Andarist:
Thank you very much @brianespinosa ❤️

PR #182, opened 2022-05-10
feat: add `labels` input to configure

NotFounds:
It would be cool to attach labels when the action creates a release PR.
For example, if we add a `release` label to PRs, we can find release PRs easily. 

I implemented `labels` input.
`labels` allows input, and we can specify comma-separated labels(include space).

e.g.
```yml
      - name: Create Release Pull Request
        uses: notfounds/changesets-action@main
        with:
          publish: yarn changeset publish
          labels: "release,test label,test-label2"
``` 
cf. https://github.com/NotFounds/changesets-monorepo/blob/d42f07d8b7def90913c71be14e9d69c8d3254876/.github/workflows/release.yml

I confirmed it works well in https://github.com/NotFounds/changesets-monorepo/pull/5.

### DIFF
--- a/.changeset/slow-actors-exercise.md
+++ b/.changeset/slow-actors-exercise.md
@@ -1,0 +1,5 @@
+---
+"@changesets/action": minor
+---
+
+Add `labels` input to configure
--- a/action.yml
+++ b/action.yml
@@ -28,6 +28,9 @@ inputs:
     description: "A boolean value to indicate whether to create Github releases after `publish` or not"
     required: false
     default: true
+  labels:
+    description: The pull request labels that comma separated. Default is empty
+    required: false
 outputs:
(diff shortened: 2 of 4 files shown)


NotFounds:
@Andarist 
Hi, Could you review this or assign a reviewer, please?

jimmynXD:
is this available yet????

Andarist:
I think this would be better as a separate action that could label PRs based on the input. This way you could provide the output of this action to that labeling action and we could include an example like that in our docs. I don't see a strong reason to include this new input here when this is easily achievable through composition.

Andarist:
I've created a PR [here](https://github.com/actions/labeler/pull/411) that should help to solve this need through composition.

PR #194, opened 2022-06-30
Change push command option to use --follow-tags instead of --tags

NotFounds:
There's no necessary to push all tags in `refs/tags`. It's enough to push only lacking tags from the remote. 
So we can use `--follow-tags` option instead of `--tags`.
cf. https://git-scm.com/docs/git-push#Documentation/git-push.txt---follow-tags
> Push all the refs that would be pushed without this option, and also push annotated tags in refs/tags that are missing from the remote but are pointing at commit-ish that are reachable from the refs being pushed.

It can prevent failing workflow with the following exception: `hint: Updates were rejected because the tag already exists in the remote.`

### DIFF
--- a/.changeset/violet-bags-sell.md
+++ b/.changeset/violet-bags-sell.md
@@ -1,0 +1,5 @@
+---
+"@changesets/action": minor
+---
+
+Change git push command option to use '--follow-tags' instead of '--tags'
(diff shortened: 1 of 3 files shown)


bluwy:
We now only explicitly pushed the tags like `git push origin <tag-name>` so I don't think this is needed anymore.

Issue #179, opened 2022-05-06
Add a requirement for git repo Settings's Workflow Permissions to the document


hcg1023:
I felt that adding a request for Workflow Permissions for Git Settings at the beginning of the document would have saved me a lot of trouble. It was my first time using Github Actions and Changesets and the constant error reporting was very annoying, The default permissions are read-only and do not allow the creation and approval of pull requests
![image](https://user-images.githubusercontent.com/47942469/167153125-5ed70d35-55aa-4aeb-9822-7d52bf6ceaf8.png)
However, you need to write and create permission to approve pull requests
![image](https://user-images.githubusercontent.com/47942469/167152932-5570d6b2-f9b0-43f5-ab67-2a1c51b68cef.png)

smiley-uriux:
Might also be useful to mention the specific permissions needed for those that can't or do not wish to grant full workflow permissions. After some experimentation, the following permissions seemed to work:
```
jobs:
  release:
    name: Prepare Release
    runs-on: ubuntu-latest
    permissions:
      id-token: write
      contents: write
      packages: write
      pull-requests: write
      issues: read
   ```

Andarist:
@hcg1023 so you have encountered a problem there because you were using a custom GitHub token with some missing scopes that are automatically included in the default token?

@smiley-uriux interesting, TIL about `jobs.{{job}}.permissions`.

I would be very much open for accepting PR that would be adding some kind of Security section to the docs.

hcg1023:
@Andarist I'm using secrets.GITHUB_TOKEN, which should not be a custom GITHUB_TOKEN

Andarist:
Ah - I see. I didn't realize that the permissions of this default token could have been changed like this. If you find a good place in the README to add this kind of information I would gladly merge a PR with such an addition

airtonix:
Tried this out: 

- have feature PR, lint/text yadda yadda
- merge feature PR
- master workflow runs: 
```yml
name: Release Package

on:
  push:
    branches:
      - master

concurrency: ${{ github.workflow }}-${{ github.ref }}

jobs:
  release:
    runs-on: ubuntu-latest
    permissions:
      id-token: write
      contents: write
      packages: write
      pull-requests: write
      issues: read
    steps:
      - uses: actions/checkout@v2
        with:
          fetch-depth: 0

      - run: git fetch --no-tags --prune --depth=5 origin master

      - name: asdf setup
        uses: asdf-vm/actions/setup@v1.1.0

[...]

PR #197, opened 2022-07-14
Clearly explain token, permission and config requirements

airtonix:
So far there seems to be a boat load of issues relating to confusion about what exactly should be used in the token and permissions.

This PR seeks to expand on the documentation in order to guide people to successful setups.


airtonix:
fixes #179

airtonix:
@Andarist I think i need some guidance on the wording of this, currently it reads like soup.

PR #233, opened 2022-11-17
Use default node version from actions/setup-node

trivikr:
Follow-up to https://github.com/changesets/action/pull/216

The actions/setup-node@v3 uses node16 by default.
The changesets/action can delegate selecting node version, and simplify configuration.

https://github.com/actions/setup-node/blob/6bc15ab23c9584a6fe2fdf1ae07fd5fb409d1dbc/action.yml#L36

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -1,7 +1,6 @@
 name: Changesets
 description: A GitHub action to automate releases with Changesets
 runs:
-  using: "node16"
   main: "dist/index.js"
 inputs:
   publish:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,12 +10,8 @@ jobs:
 
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v2
-
-      - name: Use Node.js 16
-        uses: actions/setup-node@v1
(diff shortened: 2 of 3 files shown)


action.yml
```diff
@@ -1,7 +1,6 @@
 name: Changesets
 description: A GitHub action to automate releases with Changesets
 runs:
-  using: "node16"
```
Andarist, line 4: what exact impact that would have on consumers of this action though? isn't it recommended to use an explicit version here?

Andarist:
What you have [linked to](https://github.com/actions/setup-node/blob/6bc15ab23c9584a6fe2fdf1ae07fd5fb409d1dbc/action.yml#L36) is pointing to the version used to execute `actions/setup-node` itself and not to what version of node does it setup for other things executed by the runner in steps following the setup action.

In [their README](https://github.com/actions/setup-node/tree/6bc15ab23c9584a6fe2fdf1ae07fd5fb409d1dbc#usage) we can find this:
> The node-version input is optional. If not supplied, the node version from PATH will be used. However, it is recommended to always specify Node.js version and don't rely on the system one.

PR #235, opened 2022-11-30
Respect config in .npmrc

thedv91:
Currently, this package only support to publish on `registry.npmjs.org`. It can not publish to custom registry like [Verdaccio](https://verdaccio.org/)


Andarist:
Could you provide some more information/context behind your PR? I have no idea how to start reviewing this if you don't specify what's currently broken and how this PR improves the situation.

thedv91:
@Andarist I need to publish on my custom registry but your code can only support to publish on `registry.npmjs.com`

```
http://my_ip:4873/:_authToken=$NPM_TOKEN
```

scutuatua-crypto (approved):

PR #240, opened 2022-12-06
add support for pr labels

jacklesliewise:
# Context

We would like to attach some labels to the PRs that changesets action opens.

# Changes

- Adding an optional `labels` action options
- Update the created PR with provided labels

# Considerations

- We need to use `issues` rather than `pulls` - however they mean the same thing
- Should we provide a default label? I imagine not
- Should we run `addLabel` when updating a PR as well?

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ This action for [Changesets](https://github.com/atlassian/changesets) creates a 
 - version - The command to update version, edit CHANGELOG, read and delete changesets. Default to `changeset version` if not provided
 - commit - The commit message to use. Default to `Version Packages`
 - title - The pull request title. Default to `Version Packages`
+- labels - The pull request labels. Default to none
 - setupGitUser - Sets up the git user for commits as `"github-actions[bot]"`. Default to `true`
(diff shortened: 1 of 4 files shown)


Andarist:
There is already a similar PR open [here](https://github.com/changesets/action/pull/182). I think I don't want to add this to this action as mentioned in the comment there: https://github.com/changesets/action/pull/182#issuecomment-1203054269

jacklesliewise:
Ah nice, will close this then - TIL about `actions/labeler`, hopefully your PR gets merged for that

Thanks @Andarist!

Andarist:
@jacklesliewise note that you can use it quite easily even today, just by using my action over theirs: `Andarist/actions-labeler@pr-number-input`

jacklesliewise:
@Andarist I actually managed to achieve what we wanted with the current version of `actions/labeler`

When creating the `pull_request_target` workflow you can specify a branch pattern:

```yaml
name: 'Pull Request Labeler'
on:
  pull_request_target:
    types:
      - opened
    branches:
      - 'changeset-release/**'
```

Andarist:
Nice 👍

PR #243, opened 2022-12-13
unignore dist on main branch

robertkowalski:
currently the ignore is not needed on the main branch and we are running into a bug with nektos/act which uses the go-git library.

https://github.com/go-git/go-git/issues/633

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,3 @@ node_modules
 .parcel-cache
 .cache
 *.log
-
-dist/


Andarist:
Without ignoring this we will end up with constant merge conflicts. Each PR merged will create conflicts in all pending ones. I'd really like to avoid that - it's super frustrating, especially when it happens to the community PRs

Is there any open issue in `nectos/act` related to this?

Andarist:
I would wait for that issue to be resolved first then. How do you end up with this problem in the first place though? I don't quite understand the full e2e scenario that you are facing. Is `nektos/act` fetching our main branch and **then** switching to the "requested" ref? I assume that you are using `changesets/actions@v1` or something like that and not `changesets/action@main`

PR #245, opened 2022-12-16
Support publish workflow when ignore option is set

### LINKED ISSUE #241: Publishing via action while using "ignore" option
We have a monorepo of packages and want to use the ["ignore" option](https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#ignore-array-of-packages) to skip versioning and publishing a package temporarily.

The initial run of the workflow correctly runs the version, which respects the ignore option and only consumes, bumps version, and updates changelog for packages not ignored.

However, the subsequent run of the workflow still detects the leftover changeset markdown files of the ignored package, i.e. `hasChangesets` is true and tries to run the version again. If the detected changesets are all for ignored packages, `hasChangesets` should be set to false and enter the publish flow.

From what I can tell, the "ignore" option is only read and considered in `@changesets/assemble-release-plan`.

Instead of just reading all changesets below,
https://github.com/changesets/action/blob/595655c3eae7136ff5ba18200406898904362926/src/index.ts#L36
perhaps it would be better to use `@changesets/get-release-plan` to get releasable changesets?

augustjk:
Fixes #241 

The current implementation of using `readChangesetState()` on entry is based on `@changesets/read` which simply reads all of the markdown files in `.changeset` directory.

Trigging the action while using the [`ignore` option](https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#ignore-array-of-packages) with changeset files present for both ignored and non-ignored packages, the `version` will run and consume changesets made for non-ignored packages, but leave those of ignored packages. When the PR is merged and the action is run again in this state, instead of moving on to the `publish` flow, it still enters the `version` flow and tries to make a PR, which fails as there are no releasable packages in this state.

The fix incorporates `@changesets/get-release-plan` which actually returns a list of `releases` which takes into account the ignored packages config. The flow control now utilizes this list to determine whether there's anything to version or publish.

I've done some testing of this change at https://github.com/augustjk/changesets-test-project/actions and confirmed publishing with ignored packages works.

Open questions:
- I've created the variable as `hasReleases` for the control to better reflect this, and added to the output. The `hasChangesets` variable and output no longer really serve a purpose within the code. I've left it here so as not to break users currently using this output as advertised on the README. Is this acceptable?
- I'm unsure whether this should be a minor or patch bump. Leaving the output above should keep this non-breaking, but using the count of `releases` instead of `changesets` might have some other consequences that I'm not thinking of.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "@babel/core": "^7.13.10",
     "@babel/preset-env": "^7.13.10",
     "@babel/preset-typescript": "^7.13.0",
+    "@changesets/get-release-plan": "^3.0.15",
     "@changesets/pre": "^1.0.9",
     "@changesets/read": "^0.5.3",
     "@manypkg/get-packages": "^1.1.3",
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,8 +1,8 @@
 import * as core from "@actions/core";
+import getReleasePlan from "@changesets/get-release-plan";
 import fs from "fs-extra";
 import * as gitUtils from "./gitUtils";
 import { runPublish, runVersion } from "./run";
-import readChangesetState from "./readChangesetState";
 
 const getOptionalInput = (name: string) => core.getInput(name) || undefined;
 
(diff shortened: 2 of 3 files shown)


s0:
This PR is a little old now, going to try and review / revive it to see if the fix is still valid.

Issue #262, opened 2023-02-17
question: secrets to file system


yaacovCR:
This action currently writes secrets to the file system so that they can be useful later by git/npm — would it be preferable/possible to pass the secrets via the environment ?

Andarist:
Which secrets are you referring to? IIRC we might write the node auth token to a file - but that's required by npm. I have no idea how to publish things without relying on `.npmrc`

yaacovCR:
https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages

We can use the NODE_AUTH_TOKEN environment variable that the setup node action may give us?

For git instead of netrc, we can perhaps save the credentials temporarily to memory: https://stackoverflow.com/a/36949072

yaacovCR:
Ok,

I started this issue because I was looking at the changesets action code and came upon these lines where the environment variable is directly written to the file:

https://github.com/changesets/action/blob/595655c3eae7136ff5ba18200406898904362926/src/index.ts#L76
https://github.com/changesets/action/blob/595655c3eae7136ff5ba18200406898904362926/src/index.ts#L83

What you could do instead is just what github says it does and copy the following to the .npmrc file:
`//registry.npmjs.org/:_authToken=${NODE_AUTH_TOKEN}`

I see now in the README that you reference the fact that node sets up an npmrc file with the environment variable => and of course, the code checks for the AUTH line, so obviously the code/you/someone is aware of the NPM behavior.

But I also note that the name of the environment variable you suggest using is NPM_TOKEN and not NODE_AUTH_TOKEN. Maybe that reflects an earlier/custom behavior of the setup-node action, although in terms of references, all I could find for AUTH was [this PR](https://github.com/actions/setup-node/pull/21)), and so node's automatic behavior in theory should not be working.

Presumably, this would actually cause auth errors for anyone using the action, but it seems that `setup-node` only writes the npmrc file if a `registry-url` config setting is added, and so as long as you don't add that, `changesets` would detect that no npmrc file was written, and write its own, with the NPM_TOKEN embedded in cleartext.

So I can setup a PR to change the README to suggest using NODE_AUTH_TOKEN like in the github docs, and also change the written npmrc string to use an environment variable instead of actually reading the variable and writing its value.

Let me know what I've misunderstood, I'm sure something!

After this, I could work on a PR to do something similar for netrc for the GITHUB_TOKEN....

cseas:
[...]

PR #266, opened 2023-02-26
use `NODE_AUTH_TOKEN` as environment variable name for the `NPM_TOKEN` secret

yaacovCR:
depends on #267 which was extracted from this PR

context: https://github.com/changesets/action/issues/262#issuecomment-1443370427

in a follow-up PR, one could envision a `persist`: `false` option that removes these files if the action actually created them so as to perform "clean up" similar to the `persist-credentials`: `false` option for [`actions/checkout`](https://github.com/actions/checkout#checkout-v3)

### DIFF
--- a/.changeset/brave-buckets-double.md
+++ b/.changeset/brave-buckets-double.md
@@ -1,0 +1,7 @@
+---
+"@changesets/action": major
+---
+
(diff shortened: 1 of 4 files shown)


bluwy:
We've decided to leave the npmrc handling to users or actions/checkout instead: #695

PR #267, opened 2023-02-26
write environment variable references instead of values

yaacovCR:
extracted from #266


bluwy:
We've decided to leave the npmrc handling to users or actions/setup-node instead: #695

yaacovCR:
Sounds great!

PR #279, opened 2023-04-03
Support for multiple commands in version script

RomanHotsiy:
I'll update docs and add some more error handling if you are open to accept this PR in general.

The use case for this is to run `pnpm install` after `changeset version`

### DIFF
--- a/src/run.ts
+++ b/src/run.ts
@@ -278,8 +278,11 @@ export async function runVersion({
   let versionsByDirectory = await getVersionsByDirectory(cwd);
 
   if (script) {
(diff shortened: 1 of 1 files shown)


Techn1x:
I also needed this behaviour, I suspect the intent is that `pnpm changeset version` should make the necessary changes, but it does not seem to.

I got around the 1 command limitation it by adding this to my package.json scripts
```
"scripts: {
  "version": "changeset version && pnpm install --no-frozen-lockfile",`
}
```
and then setting the version command like this
`version: pnpm run version`

Issue #192, opened 2022-06-21
403: You have exceeded a secondary rate limit


danieldelcore:
Hey folks, I seem to be running into some rate limits when running `Create Release PR or Publish to npm`

It looks my Org/Account hasn't hit a rate limit so I'm wondering if it has done so on the changesets org.

My action config looks like so: https://github.com/CodeshiftCommunity/CodeshiftCommunity/blob/main/.github/workflows/release.yml

Currently using [changesets/cli 2.6.2](https://github.com/CodeshiftCommunity/CodeshiftCommunity/blob/main/package.json#L43)

Failing build: https://github.com/CodeshiftCommunity/CodeshiftCommunity/runs/6961550448?check_suite_focus=true

<img width="870" alt="image" src="https://user-images.githubusercontent.com/3030010/174694309-709fdd7e-2094-432e-9a0b-46b9b68e8f7f.png">

akphi:
Hm, I'm not sure if this is the rate limit of your org or the rate limit of the changesets action bot, could someone clarify this?

frencojobs:
I don't really think it's possible for me to go over my personal rate limit, so I assume it's due to changesets action bot. I'm also experiencing this error, for the second time now and I only have like 4 repos and barely ran the action in like days.

Andarist:
> Hm, I'm not sure if this is the rate limit of your org or the rate limit of the changesets action bot, @Andarist do you happen to know?

To the best of my understanding it's related to [this specific token](https://github.com/CodeshiftCommunity/CodeshiftCommunity/blob/7daa9accd78917a1d5b5d026a16c5b7d7a0d8aa1/.github/workflows/release.yml#L33). Looking into the GitHub docs we can see that the primary rate limit for this kind of a token is 1000 requests / hour: https://docs.github.com/en/rest/overview/resources-in-the-rest-api#requests-from-github-actions 

However, the report here is about the "secondary rate limit" and we can read [here](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#secondary-rate-limits) that:
> Secondary rate limits are not intended to interfere with legitimate use of the API. Your normal rate limits should be the only limit you target. 

So my line of thinking is that one should always first receive an error about the primary rate limit being exceeded and only if they continue to make requests then the secondary rate limit can kick in. This doesn't quite match the provided information about this issue. 

I'm not sure what to do here because I lack the required information. Perhaps somebody could create a GitHub support ticket about this to get a better understanding of what could have caused that?

[...]

PR #286, opened 2023-05-04
fix: prevent hitting github secondary rate limits

varl:
Adds the octokit plugin for throttling / rate-limiting to fix the problem where action runs get blocked with a 403 error[1].

The `github.getOctokit`[2] function accepts a list of plugins, so this passes in the `throttling` plugin[3] to be hooked into the octokit instance.

It also needs some configuration to setup the `throttle` mechanisms, passed in to the `getOctokit` function.

[1]: https://github.com/changesets/action/issues/192
[2]: https://github.com/actions/toolkit/blob/main/packages/github/src/github.ts#LL18C40-L18C40
[3]: https://github.com/octokit/plugin-throttling.js

closes https://github.com/changesets/action/issues/192

### DIFF
--- a/.changeset/rotten-carrots-pump.md
+++ b/.changeset/rotten-carrots-pump.md
@@ -2,20 +2,9 @@
 "@changesets/action": patch
 ---
 
-add rate limit plugin for octokit
+This patch implements the [`@octokit/plugin-throttling`](https://github.com/octokit/plugin-throttling.js) plugin and [wires
+it up with the internal GitHub Octokit instance](https://github.com/actions/toolkit/tree/457303960f03375db6f033e214b9f90d79c3fe5c/packages/github#extending-the-octokit-instance).
 
-The changesets GitHub Action triggers 403 (secondary rate limits)[1]
-against the GitHub API, which causes the CI jobs to fail, and the only
-known workaround is to simply re-run the job.
+This plugin is recommended by [the Octokit docs](://octokit.github.io/rest.js/v19#throttling) as it implements all the GitHub [best practices for integrators](https://docs.github.com/en/rest/guides/best-practices-for-integrators?apiVersion=2022-11-28).
 
-This patch implements the `@octokit/plugin-throttling`[2] plugin and wires
-it up with the GitHub Octokit instance[3].
-
-This plugin is recommended by the Octokit docs[4] as it implements all
-the GitHub best practices for integrators[5].
-
-[1]: https://github.com/changesets/action/issues/192
-[2]: https://github.com/octokit/plugin-throttling.js
-[3]: https://github.com/actions/toolkit/blob/main/packages/github/src/github.ts#LL18C40-L18C40
-[4]: https://octokit.github.io/rest.js/v19#throttling
-[5]: https://docs.github.com/en/rest/guides/best-practices-for-integrators?apiVersion=2022-11-28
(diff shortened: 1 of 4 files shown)


src/run.ts
```diff
@@ -78,7 +82,38 @@ export async function runPublish({
   createGithubReleases,
   cwd = process.cwd(),
 }: PublishOptions): Promise<PublishResult> {
-  let octokit = github.getOctokit(githubToken);
+  const octokit = new (GitHub.plugin(throttling))(
+    getOctokitOptions(githubToken, {
+      throttle: {
+        onRateLimit: (retryAfter, options: any, octokit, retryCount) => {
```
Andarist, line 88: I kinda reverted the change from https://github.com/changesets/action/pull/286/commits/e2d743e2965da8982a4dbbb5d85ed3352f33e547 . The `@octokit/plugin-throttling` docs suggest that 4th parameter is available and I checked in the implementation that it's provided: https://github.com/octokit/plugin-throttling.js/blob/6289adb9c2053e322c269416c40e8402577b4958/src/index.ts#L176

src/run.ts
```diff
@@ -78,7 +82,38 @@ export async function runPublish({
...
+          octokit.log.warn(
```
Andarist, line 89: We don't use `octokit.log` anywhere else. On top of that I see that `actokit.log.info` is basically noop (can be checked [here](https://github.com/octokit/core.js/tree/928dd9fb0cb81d4798b3127b423085f4b6bfaaa6#logging)). Do u have any opinions on how logging should be done in the github actions in general and here?

Andarist: > It would be better to take a holistic pass at logging in general in a separate branch to make sure it's consistent and works as intended across the board, so I would strip out the octokit logging for now, and/or use the established pattern for logging.

This sounds like a good plan. Let's just replace the octokit logging with plain console-based functions and create a ticket to revamp logging in the future.

.changeset/rotten-carrots-pump.md
```diff
@@ -0,0 +1,21 @@
...
+it up with the GitHub Octokit instance[3].
+
+This plugin is recommended by the Octokit docs[4] as it implements all
+the GitHub best practices for integrators[5].
+
+[1]: https://github.com/changesets/action/issues/192
+[2]: https://github.com/octokit/plugin-throttling.js
+[3]: https://github.com/actions/toolkit/blob/main/packages/github/src/github.ts#LL18C40-L18C40
+[4]: https://octokit.github.io/rest.js/v19#throttling
+[5]: https://docs.github.com/en/rest/guides/best-practices-for-integrators?apiVersion=2022-11-28
```
Andarist, line 21:
```suggestion
This patch implements the [`@octokit/plugin-throttling`](https://github.com/octokit/plugin-throttling.js) plugin and [wires
it up with the internal GitHub Octokit instance](https://github.com/actions/toolkit/tree/457303960f03375db6f033e214b9f90d79c3fe5c/packages/github#extending-the-octokit-instance).

This plugin is recommended by [the Octokit docs](://octokit.github.io/rest.js/v19#throttling) as it implements all the GitHub [best practices for integrators](https://docs.github.com/en/rest/guides/best-practices-for-integrators?apiVersion=2022-11-28).

This should help with `changesets/action` gitting spurious secondary rate limits and failing CI jobs, for which the only known workaround is to simply re-run the job.
```

Andarist (approved):

PR #289, opened 2023-05-08
chore: migrate to logging from @actions/core

varl:
Use logging from `@actions/core`[1] over console-based logging recommended by the javascript action template[2].

[1]: https://github.com/actions/toolkit/tree/main/packages/core#logging
[2]: https://github.com/actions/javascript-action/blob/main/index.js

### DIFF
--- a/.changeset/blue-hornets-marry.md
+++ b/.changeset/blue-hornets-marry.md
@@ -1,0 +1,5 @@
+---
+"@changesets/action": patch
+---
+
+migrate to logging provided by @actions/core
--- a/.changeset/blue-hornets-marry.md
+++ b/.changeset/blue-hornets-marry.md
@@ -2,4 +2,4 @@
 "@changesets/action": patch
 ---
 
-migrate to logging provided by @actions/core
+Use logging provided by `@actions/core`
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -100,7 +100,7 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
       return;
     }
     case hasChangesets && !hasNonEmptyChangesets:
-      console.log("All changesets are empty; not creating PR");
+      core.warning("All changesets are empty; not creating PR");
```
Andarist, line 103:
```suggestion
      core.info("All changesets are empty; not creating PR");
```

Andarist: This is a "normal situation" so I think that it shouldn't be a warning. Happy to discuss this further if you feel differently though.

.changeset/blue-hornets-marry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/action": patch
+---
+
+migrate to logging provided by @actions/core
```
Andarist, line 5:
```suggestion
Use logging provided by `@actions/core`
```

src/index.ts
varl: No, that makes sense, thanks for the fix.

PR #291, opened 2023-05-09
fix: use throttling-wired octokit everywhere

varl:
Related to #192.

This adds the throttling plugin to other usages of octokit that were overlooked the first time around.

cc: @Andarist

### DIFF
--- a/src/run.ts
+++ b/src/run.ts
@@ -61,7 +61,7 @@ const setupOctokit = (githubToken) => {
 };
 
 const createRelease = async (
-  octokit: ReturnType<typeof github.getOctokit>,
+  octokit: ReturnType<typeof GitHub>,
   { pkg, tagName }: { pkg: Package; tagName: string }
 ) => {
   try {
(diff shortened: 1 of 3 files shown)


varl:
Ugh, test broke because the mock for `getOctokit` isn't used. Looking into a fix.

src/run.ts
```diff
@@ -26,6 +26,40 @@ import type {} from "@octokit/plugin-throttling/dist-types/types.d";
 // To avoid that, we ensure to cap the message to 60k chars.
 const MAX_CHARACTERS_PER_MESSAGE = 60000;
 
+const setupOctokit = (githubToken) => {
```
Andarist, line 29: oh, now it makes sense... I totally missed the fact that each command~ has its own octokit instance

Andarist:
@varl let me know if you need help with that

src/run.ts
varl: Same. Want me to refactor so we have a single instance for the entire module ?

varl:
@Andarist cheers, test fixed in [e70152d](https://github.com/changesets/action/pull/291/commits/e70152dfbddc3b4d3973450a31fe9b862ae35d95). If you know of a better way, feel free to change it.

Since `github.getOctokit` isn't used anymore, I refactored the type to `GitHub` from '@actions/github/lib/utils' as used here: https://github.com/actions/toolkit/blob/main/packages/github/src/github.ts#L19

Hope that works the same way as before since it's the same instance type.

PR #293, opened 2023-05-10
feat: only create the .netrc file if it doesn't already exist.

leonschiphol:
This allows custom user configuration to be used by the action without having to explicitly configure the action to support this.

### DIFF
--- a/.changeset/two-lizards-hope.md
+++ b/.changeset/two-lizards-hope.md
@@ -1,0 +1,5 @@
+---
+"@changesets/action": minor
+---
+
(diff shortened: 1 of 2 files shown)


Andarist:
Out of curiosity - what kind of a custom setup do you need there?

leonschiphol:
We want to trigger other workflows when a tag is pushed.

Andarist (approved):

Precha24 (approved):
Good, thanks

bluwy:
We've removed .netrc in v2

PR #309, opened 2023-07-01
Allow actions to specify target registry domain to publish

rayriffy:
At the moment GitHub Actions will create `.npmrc` that only point to default's NPM registry (which is `registry.npmjs.org`). This actions option will allow user to change registry domain that script will create, and configure.

### DIFF
--- a/.changeset/few-games-impress.md
+++ b/.changeset/few-games-impress.md
@@ -1,0 +1,5 @@
+---
+"@changesets/action": minor
+---
+
+allow custom registry
--- a/README.md
+++ b/README.md
@@ -122,6 +122,21 @@ For example, you can add a step before running the Changesets GitHub Action:
     NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
 ```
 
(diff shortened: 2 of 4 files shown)


ImadBoumzaoued (approved):

luisfagottani:
Hey @ImadBoumzaoued, we need this feat to be able to define a custom registry url. Can you merge this, plz!!?

Andarist:
We only create `.npmrc` if there is none already. The preferred solution to this is to include a preparation step like this one:
```ts
- name: Creating .npmrc
  run: |
    cat << EOF > "$HOME/.npmrc"
      //registry.npmjs.org/:_authToken=$NPM_TOKEN
    EOF
  env:
```

You can control your target registry in the `.npmrc` that you create.

luisfagottani:
That's right @Andarist, but I didin't need the home variable, just created with `.npmrc` path and works. 

The only thing I think it's a little strange is the output from changeset warning me that doesn't exist a npm user file, even if I created an step before.  

image

Otherwise, all good.

PR #316, opened 2023-08-02
fix: :bug: Fix tags not triggering github create event

mckelveygreg:
We've been loving changesets!

However, I think I finally tracked down an issue we were having.
Our monorepo has a cloudbuild trigger listening to tags, but github doesn't create a `Create` event [when more than 3 tags are created at once](https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads#create)... which is unfortunate when we have changes in 5 of our apps... 

I did a local test to confirm that as long as tags are pushed one at a time, they all trigger github events. 

Let me know if anything should be adjusted or if I didn't account for something!

### DIFF
--- a/.changeset/many-gifts-sort.md
+++ b/.changeset/many-gifts-sort.md
@@ -1,0 +1,5 @@
+---
+"@changesets/action": patch
+---
+
+Fix tags not triggering github create event
--- a/src/gitUtils.ts
+++ b/src/gitUtils.ts
@@ -29,8 +29,17 @@ export const push = async (
   );
 };
 
+/**
+ * Push new tags 1 by 1 to avoid github create event limits
+ */
 export const pushTags = async () => {
-  await exec("git", ["push", "origin", "--tags"]);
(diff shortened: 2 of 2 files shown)


tedraykov:
We are experiencing the same problem when we have a GitHub workflow triggered on a pushed tag. The workflow is not triggered when the changesets action pushes more than 3 tags.

mckelveygreg:
@tedraykov To unblock our CI, we released a private fork with this fix in, and reference that in our action instead: 
```yml
 - name: Create Release Pull Request
        uses: <org>/changesets-action@<tag>
```
We built it locally, so there was a `dist/index.js`, and then used that artifact as the release.

jbltx:
Also encountering the same issue. Any mono-repo will eventually face the problem.

bluwy:
We now push tags one by one since #391, and we've also switched to pushing through the GitHub CLI by default, so I think this is resolved.

Issue #260, opened 2023-02-09
`createGithubReleases` not work if I release my package to docker hub


zanminkian:
Some repos which release its output to docker hub or vscode extension marketplace use its own release script. The release script probably will not print `New tag:`. So `createGithubReleases` will not work even though I set it to `true`. This bug caused by [this line](https://github.com/changesets/action/blob/8c3f5f5637a95a2327e78d5dabcf357978aedcbb/src/run.ts#L139).

zanminkian:
I have to add `&& echo "New tag:"` at the end of the release script, which is a ugly workaround.

marcovanharten-cpi:
@zanminkian . I think that is indeed referring to the stdout that `npx changeset tag` produces: 
![image](https://github.com/changesets/action/assets/114489263/f49d1cce-b13a-45f5-91eb-a1e4f9242633)

If there is no particular reason you are not tagging in the release script, for me it did fix everything by just updating the release script as follows:
```json
{
  "release": "... && pnpm publish && changeset tag"
}
```

Hopefully this helps your workflow as well, although I think if this is the "expected behavior", they should mention this more clearly in the docs

zanminkian:
@marcovanharten-cpi Many thanks! Docs should tell use clearly.

manuel3108:
Happend to me just now. This should really be mentioned in the readme.

Andarist:
@manuel3108 PRs are welcome

manuel3108:
#345

PR #345, opened 2023-12-28
Mention required tags for option `createGithubReleases`

manuel3108:
Relates #260 
It is necessary, that the tags are created, in order for the script to create the github releases. This kinda makes sense if you think about it. The problem is that this does not seem to be included into the readme at this point, and there is no warning (or even a fail) in the cli that this would be required.

First things first, let's inform the users in the readme


manuel3108:
I've added a changeset, assuming that this will also publish the updated readme to NPM and so on

PR #349, opened 2024-01-26
Allow a custom command to run before commit

dwightjack:
This PR adds a `beforeCommit` input to allow users to run a custom command before version changes are committed and pushed to the release PR. This can be useful:

- to run additional commands after a version bump
- as an escape hatch to commit the files with different tools like https://github.com/int128/ghcp or the GitHub API (to have signed commits without the need of a bot or app).

### DIFF
--- a/.changeset/blue-fans-speak.md
+++ b/.changeset/blue-fans-speak.md
@@ -1,0 +1,5 @@
+---
+"@changesets/action": minor
+---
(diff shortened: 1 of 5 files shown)


emmatown:
You can already customise the version script which runs immediately before this new option would run so you can do whatever you need there

PR #362, opened 2024-04-10
Feat/private registry

dohaicuong:
I've encounter problem when using changesets/action when I'm using the private registry that is not npm.

Upon looking into it the problem seems to coming from the bellow line where it only check for npm registry name.
```
  // check based on https://github.com/npm/cli/blob/8f8f71e4dd5ee66b3b17888faad5a7bf6c657eed/test/lib/adduser.js#L103-L105
  return /^\s*\/\/registry\.npmjs\.org\/:[_-]authToken=/i.test(line);
});
```

However, beside npm registry, we also have github registry and custom domain gitlab registry.
```
// npm
//registry.npmjs.org/:_authToken

// github
//npm.pkg.github.com/:_authToken=TOKEN

// custom domain gitlab registry
//code.org.com/packages/npm/:_authToken=
```

Hence I'm making the change here to address for this problem.
Thank you the team for your work on maintaining this tool!

### DIFF
--- a/.changeset/nice-tips-heal.md
+++ b/.changeset/nice-tips-heal.md
@@ -1,0 +1,5 @@
+---
+"@changesets/action": minor
+---
+
+enable support for other private registry
--- a/src/index.ts
+++ b/src/index.ts
@@ -60,8 +60,10 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
         core.info("Found existing user .npmrc file");
         const userNpmrcContent = await fs.readFile(userNpmrcPath, "utf8");
         const authLine = userNpmrcContent.split("\n").find((line) => {
-          // check based on https://github.com/npm/cli/blob/8f8f71e4dd5ee66b3b17888faad5a7bf6c657eed/test/lib/adduser.js#L103-L105
-          return /^\s*\/\/registry\.npmjs\.org\/:[_-]authToken=/i.test(line);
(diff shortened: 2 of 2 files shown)


booleans-oss:
+1 as this is exactly what I need

brad-mccormick:
I think this is moving in the right direction and thanks @dohaicuong for getting this PR going. I've had the same issue since last year and I'm just now coming back to it. Our packages are stored in a private Artifactory registry which utilizes basic auth (see: [Artifactory Basic Auth](https://jfrog.com/help/r/jfrog-artifactory-documentation/authenticate-npm-using-basic-authentication)). That means the .npmrc is expected to have `username` and `_password` fields. The proposed changes would still result in authLine being false.

npm has a set of supported fields for authentication which could be mirrored. 
https://docs.npmjs.com/cli/v10/configuring-npm/npmrc#auth-related-configuration

For example:
`:[_-](authToken|auth|password)=`

There's also still some hardcoding of npm registry urls here. Maybe out of scope for this PR, but the reality is you could be publishing to more than one registry, so I'm not sure how useful this is.
https://github.com/changesets/action/blob/dec3baf00a2313782c864287697584e0eb578676/src/index.ts#L76-L87

graffhyrum:
+1

dikaso:
Also waiting for this one. How can we help?

izayl:
any progress about this pr?

PR #255, opened 2023-01-19
Feature: Allow custom `branch`

ernestognw:
Solves #254

PR #370, opened 2024-05-16
Add customize PR branch name 

cregourd:
As of previous PR #255, it's possible to trigger the Changeset Pull Request process from a branch other than `baseBranch`. Thanks for this, it's really useful  🙏

If two branches want to merge into `baseBranch`, the PR branch name will always be `changeset-release/{{baseBranch}}`. In this case, the latest commit will override the branch. 

But sometimes we just want two different branches. For example, in a Git workflow where the release can come from the `develop` branch or the `hotfixes` branch.

It should be possible to do this workflow with this PR

schema

### DIFF
--- a/.changeset/good-mice-brake.md
+++ b/.changeset/good-mice-brake.md
@@ -1,0 +1,5 @@
+---
+"@changesets/action": patch
+---
+
+Allow customize branch name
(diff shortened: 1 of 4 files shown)


oshytiko:
In big monorepos with many separate packages with different release cycles this behavior (having only one branch to create PRs from) leads to issues when changesets from different packages overwrite each others if not merged swiftly.

I think it's important to have ability to customize branch name with package name being release and version.

PR #385, opened 2024-07-31
Added makeGitHubReleasesLatest input

### LINKED ISSUE #340: Avoid marking releases as Latest
Hey,

This idea is similar to #322 and they may be considered as a single issue.

Releases created by the action do not specify `make_latest` param for the [createRelease](https://octokit.github.io/rest.js/v20#repos-create-release) call and the default behavior of Github is

> Defaults to true for newly published releases. 

With a repository being a host to many packages the Latest release doesn't necessarily mean the most recently published package. In my particular case none of the packages should claim the Latest release status. The project release workflow that we consider to be the Latest is managed separately from the changeset publishing workflow.

I'm happy to contribute with the code change or to learn that this is not the direction you envisioned for this action to take.

### Possible solution

1. Extend the action to inputs to allow set `make_latest` to `false` when a new Github release is created
2. Always create a release with `make_latest` set to `false` (potentially a breaking change to some users)

### Considered alternative

Set `createGithubReleases` to `false` and create releases outside of this action

galargh:
Resolves #340 

This PR adds a `makeGitHubReleasesLatest` input which controls whether the created GitHub releases are marked as latest or not. It defaults to `true` which is the same as the GitHub API's default value. This makes this change backwards compatible.

This addition will be useful since, especially in a monorepo setup, marking the releases of subpackges as latest is not desirable.

Please let me know if you wanted me to introduce any modifications to the approach.

### DIFF
--- a/.changeset/tiny-boxes-visit.md
+++ b/.changeset/tiny-boxes-visit.md
@@ -1,0 +1,5 @@
+---
+"@changesets/action": minor
+---
+
+Added `makeGitHubReleasesLatest` input which controls whether created GitHub releases are marked as latest or not
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ This action for [Changesets](https://github.com/atlassian/changesets) creates a 
 - title - The pull request title. Default to `Version Packages`
 - setupGitUser - Sets up the git user for commits as `"github-actions[bot]"`. Default to `true`
 - createGithubReleases - A boolean value to indicate whether to create Github releases after `publish` or not. Default to `true`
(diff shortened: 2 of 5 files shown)


galargh:
Hey! Do you think there's a chance to get this contribution incorporated?

trueberryless:
> Hey! Do you think there's a chance to get this contribution incorporated?

Unfortunately, I have been noticing that less and less PRs and issues are being worked on in this repo. Maybe the maintainers are not maintaining the changeset org any more... 😥

trueberryless:
Hi!

Just FYI: I'm no maintainer, I just reviewed the changes 👍

PR #397, opened 2024-09-04
feat: formatPrettierWithChangesets option

bennypowers:
See https://github.com/changesets/changesets/pull/1453


Andarist:
I changed https://github.com/changesets/changesets/pull/1453 to make this a config option instead of a CLI flag. So this shouldn't be needed as Changesets will just read the desired value from the config and use that automatically.

PR #415, opened 2024-10-13
Improve error message when attempting to publish without publish script defined

benmccann:
Prior to this PR it would tell you `No changesets found` if you omitted the `publish` script, and it was very hard to figure out what to do. It was also misleading as the only reason there were no changesets present is because they were deleted earlier in the release process

### DIFF
--- a/.changeset/breezy-garlics-bathe.md
+++ b/.changeset/breezy-garlics-bathe.md
@@ -1,0 +1,5 @@
+---
+"@changesets/action": patch
+---
+
+Improve error message when attempting to publish without publish script defined
--- a/src/index.ts
+++ b/src/index.ts
@@ -52,7 +52,7 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
       return;
     case !hasChangesets && hasPublishScript: {
       core.info(
-        "Attempting to publish any unpublished packages to npm"
+        "No changesets found. Attempting to publish any unpublished packages to npm"
       );
 
       let userNpmrcPath = `${process.env.HOME}/.npmrc`;
--- a/src/index.ts
+++ b/src/index.ts
@@ -48,7 +48,7 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
 
   switch (true) {
     case !hasChangesets && !hasPublishScript:
-      core.info("No publish script found");
+      core.info("No changesets present or were removed by merging release PR. Not publishing because no publish script found.");
       return;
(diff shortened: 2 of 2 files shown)


.changeset/breezy-garlics-bathe.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/action": minor
```
Andarist, line 2: IMHO the proposed change is a major change. It's a substantial change in the default behavior. If anybody is relying on publish being an opt-in then it could have a devastating effect for them if we suddenly try to publish their packages out of a blue.

emmatown: Yeah, like I'm personally using it in places where publishing is handled separately where I definitely wouldn't want this behaviour and the README documents a `Without Publishing` workflow which with this PR isn't accurate

benmccann: Ah, thanks! I'd never used this without publishing, so sorry I missed that was a way to use this. I've updated this PR to revert to the previous behavior but update the error message

src/index.ts
```diff
@@ -48,11 +48,11 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
 
   switch (true) {
     case !hasChangesets && !hasPublishScript:
-      core.info("No changesets found");
+      core.info("No publish script found");
       return;
     case !hasChangesets && hasPublishScript: {
       core.info(
-        "No changesets found, attempting to publish any unpublished packages to npm"
```
Andarist, line 55: i feel like this was indicting why an attempt to publish was made, so i think i'd revert this change

src/index.ts
```diff
@@ -48,11 +48,11 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
 
   switch (true) {
     case !hasChangesets && !hasPublishScript:
-      core.info("No changesets found");
+      core.info("No publish script found");
```
Andarist, line 51:
```suggestion
      core.info("No publish script and no changesets found");
```

.changeset/breezy-garlics-bathe.md
Andarist: We can take another look if this is he best default behavior but that will have to be done as part of the next major release

src/index.ts
benmccann: Done

Yedfree (approved):

PR #425, opened 2024-11-19
fix: #424 - Publish in custom artifactory

### LINKED ISSUE #424: Publish in custom artifactory
After spending lot of time, I looked into the code & found that, the regex pattern is always matching for only registries pointing to `npm`. In my case we have a custom artifactory the URL that looks like `//art01.spotifycloud.io`

https://github.com/changesets/action/blob/31ff97e8acd3cfd08714eab3721f55da10003255/src/index.ts#L62-L64

## Suggested solution

we can generalized the Regex and avoid hardcoding

```ts
  // Dynamically adapt to any registry by looking for :_authToken= pattern
  return /^\s*\/\/.*\/:[_-]authToken=/i.test(line);
});
```

gethari:
Fixes #424

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -3,6 +3,7 @@ import fs from "fs-extra";
 import * as gitUtils from "./gitUtils";
 import { runPublish, runVersion } from "./run";
 import readChangesetState from "./readChangesetState";
+import { extractAuthTokenLine } from "./utils";
 
 const getOptionalInput = (name: string) => core.getInput(name) || undefined;
 
@@ -59,10 +60,7 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
       if (fs.existsSync(userNpmrcPath)) {
         core.info("Found existing user .npmrc file");
         const userNpmrcContent = await fs.readFile(userNpmrcPath, "utf8");
-        const authLine = userNpmrcContent.split("\n").find((line) => {
-          // check based on https://github.com/npm/cli/blob/8f8f71e4dd5ee66b3b17888faad5a7bf6c657eed/test/lib/adduser.js#L103-L105
-          return /^\s*\/\/registry\.npmjs\.org\/:[_-]authToken=/i.test(line);
-        });
+        const authLine = extractAuthTokenLine(userNpmrcContent);
         if (authLine) {
           core.info(
             "Found existing auth token for the npm registry in the user .npmrc file"
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -105,13 +105,13 @@ export function sortTheThings(
  * @param {string} npmrcContent - The content of the .npmrc file as a string.
  * @returns {string | undefined} - The line containing the auth token or undefined if not found.
  */
-export function extractAuthTokenLine(npmrcContent:string) {
+export function extractAuthTokenLine(npmrcContent: string) {
   /**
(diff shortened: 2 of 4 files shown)


gethari:
@Andarist, could you kindly review this when you have a moment? I’m relying on this change to get my actions working and would prefer to avoid forking the repository for such a small update. Thank you!

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # @changesets/action
 
+## 1.4.10
+
+### Patch Changes
+
+- fix: publish in custom artifactories
```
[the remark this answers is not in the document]
gethari: how do i do this ?

src/index.ts
```diff
@@ -59,10 +60,7 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
       if (fs.existsSync(userNpmrcPath)) {
         core.info("Found existing user .npmrc file");
         const userNpmrcContent = await fs.readFile(userNpmrcPath, "utf8");
-        const authLine = userNpmrcContent.split("\n").find((line) => {
-          // check based on https://github.com/npm/cli/blob/8f8f71e4dd5ee66b3b17888faad5a7bf6c657eed/test/lib/adduser.js#L103-L105
-          return /^\s*\/\/registry\.npmjs\.org\/:[_-]authToken=/i.test(line);
-        });
+        const authLine = extractAuthTokenLine(userNpmrcContent);
```
Andarist, line 63: You have mentioned that the previous version of the code goes in to the `else` branch and that creates a problem. I don't understand this. Even if it goes there... it will just append a different token to ur `.npmrc` file and that's about it. Your correct token should still be used when publishing to a custom artifactory

src/utils.ts
```diff
@@ -98,3 +98,20 @@ export function sortTheThings(
...
+ */
+export function extractAuthTokenLine(npmrcContent: string) {
+  /**
+   * @see https://github.com/npm/cli/blob/8f8f71e4dd5ee66b3b17888faad5a7bf6c657eed/test/lib/adduser.js#L103-L105
+   * This regex dynamically adapts to any registry by looking for the :_auth or :_authToken= pattern.
+   */
+  const line = npmrcContent.split("\n").find((line) => {
+    return /^\s*\/\/.*\/:(_auth|_authToken)=/i.test(line); // Match both _auth and _authToken
+  });
+  return line ? line.trim() : undefined;
```
webJose, line 116:
```suggestion
  return line?.trim();
```

webJose (changes requested):

src/index.ts
mazipan-wego: No @Andarist, by creating this extract function, the code will goes to this block.

```ts
if (authLine) {
  core.info(
    "Found existing auth token for the npm registry in the user .npmrc file"
  );
}
```

and we will not create the other `.npmrc`.

src/utils.ts
mazipan-wego: This is nitpick

mazipan-wego:
This is very important fix for people that is not using default npm registry.
It's a show stopper bug.

Please help to drive this pull request.

mazipan-wego:
Duplicate w https://github.com/changesets/action/pull/362

gethari:
@Andarist do you want to consider merging this , if I resolve the conflicts ?

PR #435, opened 2025-01-10
Upgrade actions packages

enisdenjo:
Update includes the new setting output format because `set-output` is depracated and will soon be removed.


enisdenjo:
My bad, wrong repo.

Issue #441, opened 2025-02-27
[Feature] Be able to specify GitHub user override


IvanLH:
The action currently [hardcodes](https://github.com/changesets/action/blob/07dc948a9f77d5ebbdbcea0d7e452a2f57c55ca9/src/index.ts#L33) the GitHub user to be associated with the Token specified on the environment. However, if a user wanted to create the PR as a different user this would block them.

A field could be added to specify a user name override to fix this issue.

breadadams:
We're using the pre-existing `setupGitUser` option, and that is setting both the PR author _and_ the committer. Doesn't that resolve the problem?

IvanLH:
> We're using the pre-existing `setupGitUser` option, and that is setting both the PR author _and_ the committer. Doesn't that resolve the problem?

It's been a while, but I don't think so, on the related PR #443 I make the change and the GitHub user is hard coded into one of the scripts, concretely the one that creates the release PRs.

bluwy:
While I think the `.netrc` part is strange, you can already specify a different user like this: https://github.com/bombshell-dev/automation/blob/fefdda5cafd503bf3caaccd0a350c5d863d4289c/.github/workflows/publish.yml#L74-L77

I think this should resolve the issue without needing a new option. However, we should revisit `.netrc` again. (There's open issues tracking this)

PR #443, opened 2025-02-27
Added the capability to override the GitHub user

IvanLH:
Addresses Issue #441

### DIFF
--- a/.changeset/khaki-geese-speak.md
+++ b/.changeset/khaki-geese-speak.md
@@ -1,0 +1,5 @@
+---
+"@changesets/action": minor
+---
+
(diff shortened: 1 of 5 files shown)


pablogm95:
What is the status of this PR?
This is something very interesting and necessary

This adds the possibility to even make the PRs automerge by being able to specify the user and that it is not the default one, as it cannot be approve the same user who creates the PR.

@Andarist @emmatown

bluwy:
We've removed the `.netrc` part, so if you'd like to configure a different git user, use `git config --global user.name <name>` and  `git config --global user.email <email>` instead.

PR #391, opened 2024-08-25
Introduce an option to use the GitHub API to commit changes, for GPG

s0:
fixes #392

Turns out that there wasn't any good TypeScript libraries for making modifications to files directly, let alone one that looks at the files changed locally to determine what needs to be pushed to GitHub, so I went ahead and created one here: https://github.com/s0/ghcommit

I then used this repo as a test-case for the updated actions to see if it all works, and so far so good.

* [Tags are now signed by default](https://github.com/s0/ghcommit/tags):
  
  2024-08-25-183459_5620x2880_scrot

* [And commits are also signed, and attributed to github-actions](https://github.com/s0/ghcommit/pull/20/commits):
  
  2024-08-25-183708_5620x2880_scrot

* It also works with more complex situations, such as custom `version` commands, like in this repo where it auto-bumps the major version in the README: https://github.com/s0/changesets-action/pull/2

Would you like this functionality? And if so, do we want to just switch to this behavior by default + have a major version bump? or do we want to hide it behind an action input / argument for now and have it as opt-in?

My thinking is that just switching to this behavior overall makes the most sense, as most people probably want to attribute the commits to the owner of the `GITHUB_TOKEN`, and more and more people are going to require that commits are signed as the industry takes supply-chain-security more and more seriously.

PR #459, opened 2025-04-14
feat: allow using graphql API to get signed commits

### LINKED ISSUE #427: Recommendation for signing changeset PR commits
I want to be able to verify the commits made to the version PR.  Right now, my commits are unverified.  Is there a recommended strategy for signing and verifying commits?  I was looking at the documentation from GitHub [here](url).  Should I be trying to use the GitHub API to commit?

Judging by this [article](https://github.com/nautilus-cyberneering/pygithub/blob/main/docs/how_to_sign_automatic_commits_in_github_actions.md), using the github API would be one option.  It would be great if the action supported this.

adbutterfield:
Potentially closes #427 

Adds `apiProtocol` input that can be set to either `rest` (default) or `graphql` to use GitHub's graphql API to make the commit for the release PR. When used in combination with setting `setupGitUser` to `false` the commit will use the user of the provided `GITHUB_TOKEN`. 

This solves an issue we have at my company where the security team wants us to enforce signed commits on all of our repositories. With this change, the commits are signed with the service account user associated with the `GITHUB_TOKEN`.

Related change is #391 . However we can't use this approach at my company, as the security team wants us to only use our service account, and to not use GPG keys.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -31,14 +31,10 @@ inputs:
   branch:
     description: Sets the branch in which the action will run. Default to `github.ref_name` if not provided
     required: false
-  apiProtocol:
+  commitViaApi:
     description: Optionally uses graphql API. Used in combination with setting setupGitUser to false will give you signed commits.
     required: false
-    default: 'rest'
-    type: choice
-    options:
-      - rest
-      - graphql
+    default: false
 outputs:
   published:
     description: A boolean value to indicate whether a publishing is happened or not
--- a/action.yml
+++ b/action.yml
@@ -31,6 +31,14 @@ inputs:
   branch:
     description: Sets the branch in which the action will run. Default to `github.ref_name` if not provided
     required: false
+  apiProtocol:
+    description: Optionally uses graphql API. Used in combination with setting setupGitUser to false will give you signed commits.
+    required: false
+    default: 'rest'
+    type: choice
+    options:
+      - rest
+      - graphql
 outputs:
   published:
     description: A boolean value to indicate whether a publishing is happened or not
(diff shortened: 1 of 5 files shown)


adbutterfield:
Please let me know if this is something that makes sense for this project. If not, I think we'll just maintain this as a private fork. 🙇

action.yml
```diff
@@ -31,6 +31,14 @@ inputs:
...
     description: Sets the branch in which the action will run. Default to `github.ref_name` if not provided
     required: false
+  apiProtocol:
+    description: Optionally uses graphql API. Used in combination with setting setupGitUser to false will give you signed commits.
+    required: false
+    default: 'rest'
+    type: choice
+    options:
+      - rest
+      - graphql
```
adbutterfield, line 41: I guess the behavior right now isn't really `rest`. It doesn't use the API at all.

s0:
> Related change is #391 . However we can't use this approach at my company, as the security team wants us to only use our service account, and to not use GPG keys.

FTR, that mentioned PR (#391) also uses your GitHub token to commit using the API, and doesn't require you to manually set up GPG Keys. 🙂

adbutterfield:
> > Related change is #391 . However we can't use this approach at my company, as the security team wants us to only use our service account, and to not use GPG keys.
> 
> FTR, that mentioned PR (#391) also uses your GitHub token to commit using the API, and doesn't require you to manually set up GPG Keys. 🙂
> 
> (You can see linked examples that have committed using the `github-actions` user and are signed, which is only possible from the API)

(I'm a bit of a noob to all of this), but if #391 can get the job done, SGTM.

adbutterfield:
Closing as #391 got merged in.

PR #460, opened 2025-04-21
update: change double negative variable and remove unused variable

WooWan:
Update double negative variable for readability, and we don't use repo variable as we directly send `github.context.repo.repo` to octokit

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -37,8 +37,8 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
 
   let publishScript = core.getInput("publish");
   let hasChangesets = changesets.length !== 0;
-  const hasNonEmptyChangesets = changesets.some(
-    (changeset) => changeset.releases.length > 0
+  const isAllChangesetsEmpty = changesets.every(
+    (changeset) => changeset.releases.length === 0
   );
   let hasPublishScript = !!publishScript;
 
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -37,8 +37,8 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
 
   let publishScript = core.getInput("publish");
   let hasChangesets = changesets.length !== 0;
-  const hasNonEmptyChangesets = changesets.some(
```
Andarist, line 40: this wasn't broken - so let's not touch it

src/run.ts
```diff
@@ -318,7 +318,6 @@ export async function runVersion({
 }: VersionOptions): Promise<RunVersionResult> {
   const octokit = setupOctokit(githubToken);
 
-  let repo = `${github.context.repo.owner}/${github.context.repo.repo}`;
```
Andarist, line 321: I'm OK with this change - but the rest should be reverted

WooWan: I've reverted rest of it [f426667](https://github.com/changesets/action/pull/460/commits/f42666753a703990542227604ecab4852c68c48a)

src/index.ts
WooWan: Okay!

PR #467, opened 2025-05-06
Do not pre-start github API request

Vx-V:
Seems this pre-starting of request manages to time out in cases where there are a lot of changesets.
This can happen intermittently and sometimes seemingly randomly as sometimes code before last "await" will run slightly faster or slower

Possibly fixes all or many of those intermittent, large number of changesets or unknown error issues

### DIFF
--- a/.changeset/quiet-jeans-poke.md
+++ b/.changeset/quiet-jeans-poke.md
@@ -1,0 +1,5 @@
+---
+"@changesets/action": patch
+---
+
+Avoid searching for an existing pull request early.
--- a/src/run.ts
+++ b/src/run.ts
@@ -300,12 +300,6 @@ export async function runVersion({
     });
   }
 
-  const existingPullRequestsPromise = octokit.rest.pulls.list({
-    ...github.context.repo,
-    state: "open",
-    head: `${github.context.repo.owner}:${versionBranch}`,
-    base: branch,
-  });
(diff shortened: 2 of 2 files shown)


Andarist:
I don't quite understand the idea behind this pull request. We simply start the request sooner and then just "unwrap" the response that could still be pending or already resolved. How that is relevant to timeouts? We moment of `await`ing the response shouldn't have any effect on timeouts

Andarist:
Hm, but `await` isn't a request-specific mechanism - it can't notify the request object about just anything. I'll give this a thought and do some reading but I still fail to see how this would change just anything. That said - we should attach a continuation to this promise sooner in case it rejects. We don't want to end up running into an unhandledrejection. 

We can introduce this helper:
```ts
export function settled<T>(promise: Promise<T>): Promise<PromiseSettledResult<T>> {
  return promise.then(
    value => ({ status: "fulfilled", value }),
    reason => ({ status: "rejected", reason })
  );
}
```
and use it on the created `existingPullRequestsPromise` and then handle the unwrapped result appropriately after `await`ing that promise.

Andarist:
If the JS thread is meant to respond then ye - I can see that being an issue here. But at the same time... I don't see any blocking work in between changed lines 😬 

But ye, I guess this optimization really shouldn't matter all that much so I'll probably land this later

PR #472, opened 2025-05-08
Update README.md

### LINKED ISSUE #70: Ban devDeps in the root of monorepos in bolt-check

### LINKED ISSUE #70: updating Version Packages PR uses GITHUB_TOKEN instead of PAT

### LINKED ISSUE #187: PR created but action not started
You can see that the PR is opened correctly: https://github.com/unional/clibuilder/pull/375
But there is no workflow started: https://github.com/unional/clibuilder/actions

I'm setting it up using shared workflow:

```yml
jobs:
  code:
    uses: unional/.github/.github/workflows/yarn2-library-verify-cross-platform.yml@main

  release:
    uses: unional/.github/.github/workflows/yarn2-library-release.yml@main
    needs: code
    secrets: inherit
```
https://github.com/unional/clibuilder/blob/main/.github/workflows/release.yml

```yml
        id: changesets
        uses: changesets/action@v1
        with:
          version: yarn changeset version
          publish: yarn changeset publish
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
https://github.com/unional/.github/blob/main/.github/workflows/yarn2-library-release.yml

Maybe somehow the action is getting wrong information because of that?

### LINKED ISSUE #220: Changesets attributed to wrong PR, author and commit

### LINKED ISSUE #220: What pemission scopes does the action need for the generated `GITHUB_TOKEN`?

h3rmanj:
Adds information about required permission and how to trigger other workflows using this action. Tested what's working and not in https://github.com/h3rmanj/changesets-triggers.

The feature that actually makes this possible is the `commitMode: github-api` introduced in changesets/changesets#391. This just documents that it's possible.

Closes [#1545](https://github.com/changesets/action/issues/669), closes changesets/changesets#187, closes changesets/changesets#70, closes changesets/changesets#220

It probably also closes other issues that's caused by this as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,28 @@ This action for [Changesets](https://github.com/changesets/changesets) creates a
 - published - A boolean value to indicate whether a publishing has happened or not
 - publishedPackages - A JSON array to present the published packages. The format is `[{"name": "@xx/xx", "version": "1.2.0"}, {"name": "@xx/xy", "version": "0.8.9"}]`
 
+### Permissions
+
+If your repository has [restrictive default permissions](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token), you need to set the following permissions:
+
+```yaml
+permissions:
+  contents: write
+  pull-requests: write
+```
+
+You also need to check `Allow GitHub Actions to create and approve pull requests` in your repositories Actions settings.
+
+### Triggering other workflows
+
+When using the built-in `GITHUB_TOKEN`, tags or releases created by this action won't trigger other workflows. From the [GitHub Docs docs](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/triggering-a-workflow#triggering-a-workflow-from-a-workflow):
+
+> When you use the repository's `GITHUB_TOKEN` to perform tasks, events triggered by the `GITHUB_TOKEN`, will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs.
+
+To fix this, you should use a [Personal Access Token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens) or a [GitHub App token](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/making-authenticated-api-requests-with-a-github-app-in-a-github-actions-workflow) for this action. You also need to set the `commitMode` input to `github-api`.
+
+This is useful when using this action for [managing applications or non-npm packages](https://github.com/changesets/changesets/blob/main/docs/versioning-apps.md), and using tag or release triggers for custom release workflows.
+
 ### Example workflow:
 
 #### Without Publishing
--- a/README.md
+++ b/README.md
@@ -252,3 +252,11 @@ This is useful when using this action for [managing applications or non-npm pack
 When using `github-api` for the `commitMode` input, all commits and tags are GPG-signed and attributed to the user or app who owns the `GITHUB_TOKEN`.
 
 Due to calling the GitHub API more often, you may experience hitting [the rate limits](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28). This is more likely when using changesets in multiple, large, or very active projects.
+
+This mode uses [`@changesets/ghcommit`](https://github.com/changesets/ghcommit) to commit changes, which has some [known limitations](https://github.com/changesets/ghcommit?tab=readme-ov-file#known-limitations):
+
+> Due to using the GitHub API to make changes to repository contents, there are some things it's not possible to commit, and where using the Git CLI is still required.
+>
+> - Executable files
+> - Symbolic Links
+> - Submodule changes
--- a/README.md
+++ b/README.md
@@ -70,12 +70,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
-      - name: Setup Node.js 20
-        uses: actions/setup-node@v3
+      - name: Setup Node.js 22.x
+        uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version: 22.x
 
       - name: Install Dependencies
         run: yarn
@@ -106,12 +106,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
-      - name: Setup Node.js 20.x
-        uses: actions/setup-node@v3
+      - name: Setup Node.js 22.x
+        uses: actions/setup-node@v4
         with:
-          node-version: 20.x
+          node-version: 22.x
 
       - name: Install Dependencies
         run: yarn
@@ -171,12 +171,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
-      - name: Setup Node.js 20.x
-        uses: actions/setup-node@v3
+      - name: Setup Node.js 22.x
+        uses: actions/setup-node@v4
         with:
-          node-version: 20.x
+          node-version: 22.x
 
       - name: Install Dependencies
         run: yarn
@@ -215,12 +215,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
-      - name: Setup Node.js 20.x
-        uses: actions/setup-node@v3
+      - name: Setup Node.js 22.x
+        uses: actions/setup-node@v4
         with:
-          node-version: 20.x
+          node-version: 22.x
 
       - name: Install Dependencies
         run: yarn
--- a/README.md
+++ b/README.md
@@ -38,16 +38,6 @@ permissions:
 
 You also need to check `Allow GitHub Actions to create and approve pull requests` in your repositories Actions settings.
 
-### Triggering other workflows
-
-When using the built-in `GITHUB_TOKEN`, tags, releases and some pull request events created by this action won't trigger other workflows. From the [GitHub Docs docs](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/triggering-a-workflow#triggering-a-workflow-from-a-workflow):
-
-> When you use the repository's `GITHUB_TOKEN` to perform tasks, events triggered by the `GITHUB_TOKEN`, will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs.
-
-To fix this, you should use a [Personal Access Token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens) or a [GitHub App token](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/making-authenticated-api-requests-with-a-github-app-in-a-github-actions-workflow) for this action. You also need to set the `commitMode` input to `github-api`.
-
-This is useful when using this action for [managing applications or non-npm packages](https://github.com/changesets/changesets/blob/main/docs/versioning-apps.md), and using tag or release triggers for custom release workflows.
-
 ### Example workflow:
 
 #### Without Publishing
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,6 +20,28 @@ This action for [Changesets](https://github.com/changesets/changesets) creates a
 - published - A boolean value to indicate whether a publishing has happened or not
 - publishedPackages - A JSON array to present the published packages. The format is `[{"name": "@xx/xx", "version": "1.2.0"}, {"name": "@xx/xy", "version": "0.8.9"}]`
 
+### Permissions
```
Andarist, line 23: It might be worth mentioning that when publishing with provenance, you'll need `id-token: write` too. An example of how this can be configured: https://github.com/changesets/changesets/pull/1659/files 

Note there are other ways to enable this - but they all require this permission.

README.md
```diff
@@ -20,6 +20,28 @@ This action for [Changesets](https://github.com/changesets/changesets) creates a
...
+
+You also need to check `Allow GitHub Actions to create and approve pull requests` in your repositories Actions settings.
+
+### Triggering other workflows
+
+When using the built-in `GITHUB_TOKEN`, tags, releases and some pull request events created by this action won't trigger other workflows. From the [GitHub Docs docs](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/triggering-a-workflow#triggering-a-workflow-from-a-workflow):
+
+> When you use the repository's `GITHUB_TOKEN` to perform tasks, events triggered by the `GITHUB_TOKEN`, will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs.
+
+To fix this, you should use a [Personal Access Token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens) or a [GitHub App token](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/making-authenticated-api-requests-with-a-github-app-in-a-github-actions-workflow) for this action. You also need to set the `commitMode` input to `github-api`.
```
Andarist, line 41: Hm, IIRC some of the restrictions are lifted when using PAT - even with the default `commitMode`. Do you know what's the breakdown of what becomes possible with PAT using both modes?

Andarist, line 41: We should also add a section of the `githup-api` mode and link to it from here. I think we should mention there it has some limitations (cc @s0), that it's creating signed commits and that it might increase the likelihood of running into rate limits (I have not verified that but it only makes sense to assume that).

h3rmanj: Ah yes, I'll update some of the info.

But the default mode will create releases and pull requests with the API, which will trigger custom workflows with a PAT.

But it will use `git` commands to push tags, and update PRs, which won't trigger workflows, unless you specify the PAT in the for the `actions/checkout` i think.

h3rmanj: I think we could reduce the API requests when creating releases/tags in half, if you provide [`target_commitish`](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release) to the [`createRelease` call](https://github.com/changesets/action/blob/main/src/run.ts#L44), and drop the [`createRef` call](https://github.com/changesets/action/blob/main/src/git.ts#L66).

> Specifies the commitish value that determines where the Git tag is created from. Can be any branch or commit SHA. Unused if the Git tag already exists. Default: the repository's default branch.

README.md
h3rmanj: I can add that in as an optional permission

README.md
s0: I think it makes sense to highlight the impact on API usage on that commitMode definitely. 👍

README.md
```diff
@@ -20,6 +20,30 @@ This action for [Changesets](https://github.com/changesets/changesets) creates a
...
+### Permissions
+
+If your repository has [restrictive default permissions](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token), you need to set the following permissions:
+
+```yaml
+permissions:
+  contents: write
+  pull-requests: write
+  # uncomment this if you're publishing with provenance https://docs.npmjs.com/generating-provenance-statements
+  # id-token: write
```
s0, line 32: Nice! 👍

README.md
s0: Yes exactly. It won't always be necessary to set `commitMode` to `github-api` and use a different `GITHUB_TOKEN` for workflows to be triggered, (though IMO it is generally a good idea because it will work in all scenarios, and signed commits are good).

For the avoidance of doubt:

- `push` - workflows triggered by `push` will not run when committed to using `git-cli`, as the `git push` itself is done by GitHub Actions, and so is blocked to prevent recursion. (to address this, you must use `commitMode`: github-api`
- `pull_request` - workflows triggered by a PR event (e.g. `opened` / `synchronize`) will still continue to run, regardless of `commitMode`, as long as the `GITHUB_TOKEN` is for something other than GitHub Actions.
- `release` - this one I'm a little unsure on, I think it is only created via the API already, and so changing `GITHUB_TOKEN` is probably sufficient.

So I only think that `push` workflows are affected.

That being said, I agree that it's probably a good idea to just always recommend it, with the rate-limiting caveat. How about this:

```suggestion
To fix this, you should use a [Personal Access Token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens) or a [GitHub App token](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/making-authenticated-api-requests-with-a-github-app-in-a-github-actions-workflow) for this action.
For some events (e.g. `push`) you will also need to set the `commitMode` input to `github-api` for workflows to be correctly triggered,
however be aware that doing so may have an impact on your GitHub API rate-limits. 
```

h3rmanj: `tags` as well, but that's a subset of `push`. I also think `pull_request.synchronize` won't be triggered, because that comes indirectly from `push`. I'll do some testing and report back

h3rmanj: Full breakdown from some testing

| Variation                    | `push.branches` | `push.tags` | `pull_request.opened` |   `pull_request.synchronized`   | `release.published` |
| :--------------------------- | :-------------: | :---------: | :-------------------: | :-----------------------------: | :-----------------: |
| Built-in token               |       ❌        |     ❌      |          ❌           |               ❌                |         ❌          |
| PAT                          |       ❌        |     ❌      |          ✅           |               ❌                |         ✅          |
| PAT + `github-api`           |       ✅        |     ✅      |          ✅           | ❌ re-triggers `opened` instead |         ✅          |
| PAT + `actions/checkout` PAT |       ✅        |     ✅      |          ✅           |               ✅                |         ✅          |

Not sure if I should document all these possible variations though. Personally I prefer `github-api` mode for the following reasons:

- Commit signing
- It's built-in to the action itself, so you can achieve desired result by only configuring this action
- Once configured, it enables all event triggers, not just some

h3rmanj: Will do. I did however notice some possible optimizations for `github-api` we could explore as well.

Firstly, creating the tag implicitly through the release request. We'd have to do some testing if that still achieves desired tag signing.

Second, I noticed that when it opens a new pull request:

- It first pushes the base branch to the changeset branch <- This could be removed
- Then commits the result from version
- Then opens a pull request

And on subsequent changes, it also closes the pull request before opening a new one, which I suspect could be removed as well.

I can open a separate issue on that though.

s0: > It first pushes the base branch to the changeset branch <- This could be removed

It just occurs to me that this is why the PRs become closed, because base will match the head, and why the PRs need to be reopened.

Not sure it's possible though, from what I recall the GraphQL mutation to add files doesn't allow you to specify a different base ref to what's currently on the branch (which is why resetting it is potentially necessary), but I may be misremembering.

README.md
```diff
@@ -6,19 +6,37 @@ This action for [Changesets](https://github.com/changesets/changesets) creates a
...
+| :--------------------- | :------------------ | :-------------------------------------------------------------------------------------------------------------------------------------------- |
+| `publish`              |                     | The command to use to build and publish packages                                                                                              |
+| `version`              | `changeset version` | The command to update version, edit CHANGELOG, read and delete changesets                                                                     |
+| `commit`               | `Version Packages`  | The commit message to use                                                                                                                     |
+| `title`                | `Version Packages`  | The pull request title                                                                                                                        |
+| `setupGitUser`         | `true`              | Sets up the git user for commits as `"github-actions[bot]"`                                                                                   |
```
s0, line 15: Maybe good to note here that it doesn't do anything when `commitMode` is `github-api`? 🤔

h3rmanj:
I restructured the readme a bit. I changed the inputs and outputs section to use tables instead, can revert if that's not desirable.

Added an advanced section as well, after examples, with `Triggering other workflows` and `GitHub API commit mode` sections. Not sure if I should mention anything else in the github-api section, feel free to add anything there 😄

s0:
> Not sure if I should mention anything else in the github-api section, feel free to add anything there 😄

Probably worthwhile adding a bit about its limitations, namely that it won't commit symlinks, executable files or submodules (as this isn't supported by the API). We should probably add a section to the changesets/ghcommit README too, and link to that from this README

README.md
```diff
@@ -216,3 +234,21 @@ If you are using [Yarn Plug'n'Play](https://yarnpkg.com/features/pnp), you shoul
...
+
+When using the built-in `GITHUB_TOKEN`, tags, releases and pull requests created by this action won't trigger other workflows. From the [GitHub Docs docs](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/triggering-a-workflow#triggering-a-workflow-from-a-workflow):
+
+> When you use the repository's `GITHUB_TOKEN` to perform tasks, events triggered by the `GITHUB_TOKEN`, will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs.
+
+To fix this, you should use a [Personal Access Token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens) or a [GitHub App token](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/making-authenticated-api-requests-with-a-github-app-in-a-github-actions-workflow) for this action. You also need to set the `commitMode` input to `github-api`.
+
+This is useful when using this action for [managing applications or non-npm packages](https://github.com/changesets/changesets/blob/main/docs/versioning-apps.md), and using tag or release triggers for custom release workflows.
```
h3rmanj, line 248: > Hm, IIRC some of the restrictions are lifted when using PAT - even with the default commitMode. Do you know what's the breakdown of what becomes possible with PAT using both modes?

Full breakdown from some testing

| Variation                    | `push.branches` | `push.tags` | `pull_request.opened` |   `pull_request.synchronized`   | `release.published` |
| :--------------------------- | :-------------: | :---------: | :-------------------: | :-----------------------------: | :-----------------: |
| Built-in token               |       ❌        |     ❌      |          ❌           |               ❌                |         ❌          |
| PAT                          |       ❌        |     ❌      |          ✅           |               ❌                |         ✅          |
| PAT + `github-api`           |       ✅        |     ✅      |          ✅           | ❌ re-triggers `opened` instead |         ✅          |
| PAT + `actions/checkout` PAT |       ✅        |     ✅      |          ✅           |               ✅                |         ✅          |

Not sure if I should document all these possible variations though. Personally I prefer `github-api` mode for the following reasons:

- Commit signing
- It's built-in to the action itself, so you can achieve desired result by only configuring this action
- Once configured, it enables all event triggers, not just some

@Andarist what would you like to be mentioned here?

s0:
@h3rmanj the https://github.com/changesets/ghcommit README has a section on known limitations now if you want to link to that?

PR #473, opened 2025-05-08
Make git add work consistently with subdirectories

s0:
This is a follow-up from #471

Currently, depending on your `commitMode`, when running the action with `cwd` set to a subdirectory of a repo, the action will either:

- `git-cli`: add the files changed in the subdirectory only
- `github-api`: add files changed anywhere in the repository

This commit brings github-api in line with git-cli, so that it only adds files in the current working directory.

Related: https://github.com/changesets/ghcommit/issues/29

### DIFF
--- a/src/git.ts
+++ b/src/git.ts
@@ -4,11 +4,11 @@ import * as github from "@actions/github";
 import { commitChangesFromRepo } from "@changesets/ghcommit/git";
 import { Octokit } from "./octokit";
 
-type ExecOptions = {
+type GitOptions = {
   cwd: string;
-}
+};
 
-const push = async (branch: string, options: ExecOptions) => {
+const push = async (branch: string, options: GitOptions) => {
   await exec(
     "git",
     ["push", "origin", `HEAD:${branch}`, "--force"].filter<string>(
@@ -20,7 +20,7 @@ const push = async (branch: string, options: ExecOptions) => {
 
 const switchToMaybeExistingBranch = async (
   branch: string,
-  options: ExecOptions
+  options: GitOptions
 ) => {
   let { stderr } = await getExecOutput("git", ["checkout", branch], {
     ignoreReturnCode: true,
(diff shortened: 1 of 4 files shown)


src/git.ts
```diff
@@ -87,6 +87,13 @@ export class Git {
 
   async pushChanges({ branch, message }: { branch: string; message: string }) {
     if (this.octokit) {
+      /** 
+       * Only add files form the current working directory
+       * 
+       * This will emulate the behavior of `git add .`,
+       * used in {@link commitAll}.
+       */
+      const addFromDirectory = process.cwd();
```
Andarist, line 96: We should add `cwd` as an option in the constructor and perhaps even pass it explicitly to all `exec` calls.

s0: Would this be instead of or in addition to calling `process.chdir()` here: 

https://github.com/changesets/action/blob/main/src/index.ts#L18-L22

Given that the currently released version has this discrepancy with usage of `git add`, I'd be inclined to do that refactor in a follow-up pull request? Just to limit risk of introducing more bugs with this bugfix PR :) What do you think?

s0: Ah okay cool, yeah I was thinking if we were to also remove the call to `chdir()` in the same PR. I'll make those changes shortly then! 👍

s0: Opened #475 to track removal of `process.chdir()`, there will need to be some work in `ghcommit` to account for some edge cases.

src/git.ts
```diff
@@ -1,63 +1,80 @@
 import * as core from "@actions/core";
-import { exec, getExecOutput } from "@actions/exec";
+import { exec, ExecOptions, getExecOutput } from "@actions/exec";
```
Andarist, line 2: I'd stay away from reusing `ExecOptions` here. We want a new set of options compatible with both strategies we are using here. As of now, we'd only support `cwd` - something that is common to both. But I have doubt the github API mode would support everything `ExecOptions` has today

s0: Defined a new type for this now, also makes `cwd` required, ensuring that we're passing it everywhere

Andarist (approved):

s0:
@Andarist I can't merge this PR, so happy for you to press the button!

PR #478, opened 2025-05-10
Run typecheck on CI

Andarist:


s0 (approved):
Nice! 👍

s0:
Perhaps we should also get prettier and eslint set up here too? (In different PRs)

Andarist:
yeah, that would also be nice to have :)

PR #481, opened 2025-05-11
Removed `fs-extra` dependency

Andarist:


src/index.ts
```diff
@@ -1,5 +1,5 @@
 import * as core from "@actions/core";
-import fs from "fs-extra";
+import fs from "node:fs";
```
s0, line 2: Synchronous / Blocking FS calls should always be avoided, you can import the promises object which includes async fs calls and use it directly:

```suggestion
import {promises as fs} from "node:fs";
```

s0 (changes requested):

src/index.ts
Andarist: ye, right - this action is really only executed alone though and its flow is really linear *anyway*, so using sync variants felt just simpler. I switched back everything to `fs/promises` now though

PR #484, opened 2025-05-11
Migrate to ESM

Andarist:


.node-version
```diff
@@ -0,0 +1 @@
+v22.9.0
```
Andarist, line 1: I am fuzzy on this. This repo only has to support one node version. I had issues using `--experimental-strip-types` on node 20 so I decided to use node 22 here - for dev workflows and on CI. But then only node 20 is currently supported for GitHub Actions (`using: 'node22'` doesn't work yet).

package.json
```diff
@@ -18,13 +19,13 @@
     "vitest": "^3.1.3"
   },
   "scripts": {
-    "build": "esbuild src/index.ts --bundle --platform=node --target=node20 --minify --outfile=dist/index.js",
+    "build": "esbuild src/index.ts --bundle --platform=node --target=node20 --format=esm --minify --outfile=dist/index.js",
```
Andarist, line 22: In theory... this should be quite safe but then... who knows. Perhaps this should only land after testing this e2e in some repo. It feels like a somewhat risky change.

src/run.ts
```diff
@@ -286,9 +292,21 @@ export async function runVersion({
...
-    });
+    await exec(
+      "node",
+      [
+        fileURLToPath(
+          resolve(
+            "@changesets/cli/bin.js",
+            pathToFileURL(path.join(cwd, "x.cjs")).toString()
+          )
+        ),
```
Andarist, line 303: this is... quite a dance to make it work. Not the biggest issue but I wish there would be a builtin way to make this easier. 

If I'm not mistaken, it also means that a bug was introduced here:
https://github.com/changesets/changesets/blob/268a29fedc948f22c672a3b1e3e51df4427f478d/packages/cli/src/commit/getCommitFunctions.ts#L16

A directory can't be used directly as input to `resolve`. It should be a filepath. cc @bluwy for double-checking my logic

scripts/bump.ts
```diff
@@ -0,0 +1,18 @@
+import { exec } from "@actions/exec";
+import fs from "node:fs";
+import path from "node:path";
+import pkgJson from "../package.json" with { type: "json" };
+
+process.chdir(path.join(import.meta.dirname, ".."));
```
Andarist, line 6: We only use `import.meta.dirname` in tests and script so far so it's fine anyway. But I wonder if there are any compatibility gotchas to consider if we'd use this in the code of the action itself.

I'm not sure what is the caveat the docs mention [here](https://nodejs.org/api/esm.html#importmetadirname):
> Caveat: only present on file: modules.

Is that only that it wouldn't be present for modules imported/downloaded using HTTP(s) etc?

src/run.ts
```diff
@@ -162,7 +166,9 @@ export async function runPublish({
 
 const requireChangesetsCliPkgJson = (cwd: string) => {
   try {
-    return require(resolveFrom(cwd, "@changesets/cli/package.json"));
+    return require(require.resolve("@changesets/cli/package.json", {
```
Andarist, line 169: I'm not sure why we were using `resolveFrom`. `require.resolve` supported this for a long time already. It also just seems like using `require` here is an easier option than dancing around this using `import-meta-resolve`.

src/run.ts
bluwy: Yeah I think you're right. That should probably be a file path otherwise it resolves from one directory higher, which probably coincidentally worked because there's no `node_modules` in `.changeset/` for the linked code.

About this code, could we also use `require.resolve` instead of `import-meta-resolve` too?

scripts/bump.ts
bluwy: Yeah I think that's what it means. On GitHub workflows the code should be ran as files though and shouldn't have this issue, though you need to account for it resolving from `dist` instead.

Andarist: > though you need to account for it resolving from dist instead.

Good point. It's not an issue right now as we don't use it in `src` (and all files are nested directly in `src` too). But I'll keep that in mind as something to be more careful about.

src/run.ts
Andarist: > About this code, could we also use require.resolve instead of import-meta-resolve too?

Yep, that was my plan - I just wanted to have a pushed out reference of how to do this using `import-meta-resolve` ;p I simplified this now.

package.json
Andarist: Damn. That's not great :p I think I'll still try to run with it - but I have to definitely test it e2e first

.node-version
kristentr: thank you about it

PR #485, opened 2025-05-12
fix: create tags implicitly through release api

### LINKED ISSUE #474: Possible optimizations for `commitMode: github-api`
When doing my testing in https://github.com/changesets/action/pull/472, I noticed some possible oppertunities for optimization for the `commitMode: github-api`.

**Creating tags/releases**

https://github.com/changesets/action/pull/472#discussion_r2079844415
> I think we could reduce the API requests when creating releases/tags in half, if you provide [target_commitish](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release) to the [createRelease call](https://github.com/changesets/action/blob/main/src/run.ts#L44), and drop the [createRef call](https://github.com/changesets/action/blob/main/src/git.ts#L66) that creates the tag.

We would need to check that it still achieves the desired results in regards to signing tags.

**Creating pull requests**

https://github.com/changesets/action/pull/472#discussion_r2081201212

When `commitMode: github` creates a pull request, it does the following:

- It first pushes the base branch to the changeset branch <- This could be removed
- Then commits the result from version
- Then opens a pull request

And on subsequent changes, it also closes the pull request before opening a new one (in addition to the double push), which I suspect could be removed as well.

cc @s0

h3rmanj:
Uses `target_commitish` to create a tag when it doesn't exist.

From https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release
> Specifies the commitish value that determines where the Git tag is created from. Can be any branch or commit SHA. Unused if the Git tag already exists. Default: the repository's default branch.

closes #474

### DIFF
--- a/.changeset/nervous-eggs-bathe.md
+++ b/.changeset/nervous-eggs-bathe.md
@@ -1,0 +1,5 @@
+---
+"@changesets/action": minor
+---
+
+Reduce the number of API calls when releasing with `commitMode: github-api`
--- a/src/git.ts
+++ b/src/git.ts
@@ -77,16 +77,8 @@ export class Git {
 
   async pushTag(tag: string) {
     if (this.octokit) {
-      return this.octokit.rest.git
-        .createRef({
-          ...github.context.repo,
-          ref: `refs/tags/${tag}`,
-          sha: github.context.sha,
-        })
-        .catch((err) => {
-          // Assuming tag was manually pushed in custom publish script
-          core.warning(`Failed to create tag ${tag}: ${err.message}`);
-        });
+      // a tag will be created automatically when creating a release
+      return;
     }
     await exec("git", ["push", "origin", tag], { cwd: this.cwd });
   }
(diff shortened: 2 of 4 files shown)


h3rmanj:
> Unused if the Git tag already exists.

They will be auto-created if the tag doesn't exist. However `changesets/action` pushes the tag manually first in both `git-cli` and `github-api` commit modes, so unless those calls have been circumvented somehow (which it seems they did in the issue you mentioned), the tags wont be auto-created with the releases API.

This PR is also mainly for the `github-api` commit mode, where we can drop the explicit call to create a tag first, and instead have it auto-created by the release API.

h3rmanj:
I'm going to write a check that makes this only pass the SHA for `target_commitish` to the release API if the commit mode is `github-api`, to prevent any potential side-effects for `git-cli` mode.

Andarist:
@s0 do u have any thoughts on this one?

PR #486, opened 2025-05-12
fix: use latest ghcommit and fix cwd for git-cli

### LINKED ISSUE #475: Remove usage of `process.chdir()`
See: https://github.com/changesets/action/pull/473#discussion_r2081314357

As of #473, we now explicitly set the `cwd` argument for all subprocess calls. As such, it shouldn't be necessary to change the working directory of the current process.

We need to ensure that usage of `commitChangesFromRepo` (which makes a call to `process.cwd()`) also works correctly for our use-cases before we can remove this process call.

Namely, we need a way for running this action to work when:

- a repository is checked-out into a sub-directory.
- `cwd` is set to a subdirectory of that subdirectory

In particular, we should:

* Update `commitChangesFromRepo` so that `repoDirectory` can be used to recursively start a search for the root of a repository.
* Specify `repoDirectory` to be explicit `cwd`, as with `addFromDirectory`.

s0:
- `cwd` was not working correctly when commitMode was set to `git-cli`, as it was attempting to resolve the `cwd` twice. (This problem only occured when specifying a relative path for `cwd`.)
- Update `ghcommit`, and avoid calling `process.chdir()` to change the working directory. Instead, we now pass the `cwd` option everywhere.

fixes #475

### DIFF
--- a/.changeset/lemon-garlics-sell.md
+++ b/.changeset/lemon-garlics-sell.md
@@ -1,0 +1,9 @@
+---
+"@changesets/action": patch
+---
+
+fix: cwd now works again with relative paths and commitMode: git-cli
+
+v1.5.2 introduced a bug where specifying `cwd`
+when using the default `commitMode` would break things,
+as it tried to resolve the path twice.
--- a/.changeset/lemon-garlics-sell.md
+++ b/.changeset/lemon-garlics-sell.md
@@ -2,8 +2,4 @@
 "@changesets/action": patch
 ---
 
-fix: cwd now works again with relative paths and commitMode: git-cli
-
-v1.5.2 introduced a bug where specifying `cwd`
-when using the default `commitMode` would break things,
-as it tried to resolve the path twice.
+Fixed situations in which `cwd` was specified as a relative path and used with (default) `commitMode: git-cli`
--- a/.changeset/sour-elephants-switch.md
+++ b/.changeset/sour-elephants-switch.md
@@ -1,7 +1,2 @@
 ---
-"@changesets/action": patch
 ---
-
-Update @changesets/ghcommit, and avoid calling `process.chdir()`
-
-fixes #475
--- a/.changeset/sour-elephants-switch.md
+++ b/.changeset/sour-elephants-switch.md
@@ -1,0 +1,7 @@
+---
+"@changesets/action": patch
+---
+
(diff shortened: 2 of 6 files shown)


s0:
@Andarist please take a look and review.

src/index.ts
```diff
@@ -16,12 +17,7 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
     return;
   }
 
-  const inputCwd = getOptionalInput("cwd");
-  if (inputCwd) {
-    core.info("changing directory to the one given as the input");
-    process.chdir(inputCwd);
-  }
-  const cwd = inputCwd || process.cwd();
+  const cwd = path.resolve(getOptionalInput("cwd") ?? "");
```
Andarist, line 20: @s0 I have simplified this slightly, this should still be equivalent, right? `path.resolve` already takes `process.cwd()` into account by default

s0:
```
~ > node
> require('path').resolve('')
'/home/sam'
```

LGTM!

s0 (commented):
Changes LGTM

.changeset/lemon-garlics-sell.md
```diff
@@ -0,0 +1,9 @@
+---
+"@changesets/action": patch
+---
+
+fix: cwd now works again with relative paths and commitMode: git-cli
```
Andarist, line 5:
```suggestion
`cwd` now works again with relative paths and `commitMode: git-cli`
```

.changeset/lemon-garlics-sell.md
```diff
@@ -0,0 +1,9 @@
...
+
+v1.5.2 introduced a bug where specifying `cwd`
+when using the default `commitMode` would break things,
+as it tried to resolve the path twice.
```
Andarist, line 9:
```suggestion
```

.changeset/sour-elephants-switch.md
```diff
@@ -0,0 +1,7 @@
+---
+"@changesets/action": patch
+---
+
+Update @changesets/ghcommit, and avoid calling `process.chdir()`
+
+fixes #475
```
Andarist, line 7:
```suggestion
---
---
```

oshytiko:
I think this fix uncovered another issue, could someone take a look at [this PR](https://github.com/changesets/action/pull/502) please!

ping @s0 @Andarist

PR #497, opened 2025-05-17
Update markdown-related packages

Andarist:
cc @remcohaszing, maybe you'd have a sec to double-check if I'm using the most kosher tools here? 😅

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,10 +1,8 @@
-import unified from "unified";
+import { unified } from "unified";
 import remarkParse from "remark-parse";
 import remarkStringify from "remark-stringify";
 import fs from "node:fs/promises";
-import type { Root } from "mdast";
-// @ts-ignore
-import mdastToString from "mdast-util-to-string";
+import { toString as mdastToString } from "mdast-util-to-string";
 import { getPackages, type Package } from "@manypkg/get-packages";
 
 export const BumpLevels = {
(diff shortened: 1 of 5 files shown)


src/utils.ts
```diff
@@ -37,7 +35,7 @@ export async function getChangedPackages(
 }
 
 export function getChangelogEntry(changelog: string, version: string) {
-  let ast = unified().use(remarkParse).parse(changelog) as Root;
+  let ast = unified().use(remarkParse).parse(changelog);
```
remcohaszing, line 38: Should you support GFM and/or frontmatter here? Is so, you should use [`remark-gfm`](https://github.com/remarkjs/remark-gfm) / [`remark-frontmatter`](https://github.com/remarkjs/remark-frontmatter).

Unless you need options passed into the function, I recommend to define the processor once outside of the function. This way plugins can cache some of the heavy lifting. This isn’t really applicable to your case, but IMO it’s a good practice. If you use GFM/frontmatter, this also allows you to keep the processor in sync with the stringify processor you use below.

```ts
const processor = unified()
  .use(remarkParse)
  .use(remarkFrontmatter)
  .use(remarkGfm)
  .use(remarkStringify)

// …

let ast = processor.parse(changelog)
```

If you don’t want any of that, the remark abstractions don’t really offer much. You might as well use the slightly lower level [`mdast-util-from-markdown`](https://github.com/syntax-tree/mdast-util-from-markdown) and [`mdast-util-to-markdown`](https://github.com/syntax-tree/mdast-util-to-markdown).

remcohaszing (approved):
I added some tips, but really the tools you’re currently using are fine as-is too. :)

src/utils.ts
Andarist: Thanks for the review! ❤️

bluwy:
Closing this for now as we've refactored this away, or we can open a new pr from scratch again.

PR #499, opened 2025-05-18
Add missing `outputs` in documentation

StyleShit:
I needed to get the PR number, according the the docs it's not exposed, but after looking at the code I found out that it is.

I assumed it would be nice to include it in the docs for the next person

maybe we can add some kind of automation for that


StyleShit:
bump

PR #514, opened 2025-07-23
changesets/action support Sparse Checkout

alex-page:
## The Problem

The changesets/action fails with "There is no .changeset directory in this project" when using sparse checkout, even when the `cwd` parameter is specified. I am using sparse checkout to take a folder from a very large GitHub repository and this folder has it's own `.changeset` directory.

## Root Cause

The error occurs during the initial validation phase of the action, before any custom commands run. Here's the code flow:

1: In changesets/action index.ts we call `let { changesets } = await readChangesetState();` without passing the `cwd` in.

https://github.com/changesets/action/blob/04d574e831923498156e0b2b93152878063203a3/src/index.ts#L46

2: This is passed into `readChangesetState.ts` where the `cwd` is set to `process.cwd()` because the `cwd` from the action is not passed through.

https://github.com/changesets/action/blob/04d574e831923498156e0b2b93152878063203a3/src/readChangesetState.ts#L14

3. **The `readChangesets` function from `@changesets/read`** looks for `.changeset` at the provided path and throws the error when not found.

https://github.com/changesets/changesets/blob/c7b6832a7a2783073e720d2085a546810e9b55eb/packages/read/src/index.ts#L22-L35

### Why This Happens
- `index.ts` calls `readChangesetState()` without the `cwd` parameter
- `readChangesetState` defaults to `process.cwd()` (repository root)
- With sparse checkout, `.changeset` doesn't exist at the repo root
- The action fails during initialization, before `version` or `publish` commands run


dist/index.js
alex-page, line 1: This will be removed. Using it to test the action pointed to this commit.

PR #545, opened 2025-11-24
fix: conditionally append NPM_TOKEN to .npmrc for trusted publishing …

ryanbas21:
I ran into a bit of a headache with trusted publishing with this action.

Because trusted publishing removes the need for an npm token, and is now the more preferred approach, we should have a little smarter generation of the .npmrc file.

If we don't have a .npmrc file, and we don't have an npm token, we shouldn't automatically append an npm token to the url.

If we have an npm token in the environment, we will use it, as was the default previously.

This way, if there is no npmrc file, we dont try to use an undefined token, we just assume authentication. if they don't provide an npm token, and don't have an npm token in the environment, publishing should fail on the npm side.

### DIFF
--- a/.changeset/npmrc-trusted-publishing.md
+++ b/.changeset/npmrc-trusted-publishing.md
@@ -2,6 +2,4 @@
 "@changesets/action": patch
 ---
 
-fix: conditionally append NPM_TOKEN to .npmrc for trusted publishing support
-
-The .npmrc generation now intelligently handles both traditional NPM token authentication and trusted publishing scenarios by only appending the auth token when NPM_TOKEN is defined. This prevents 'undefined' from being written to the registry configuration when using OIDC tokens from GitHub Actions trusted publishing.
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -79,20 +81,28 @@ const getOptionalInput = (name: string) => core.getInput(name) || undefined;
...
+          );
+        } else {
+          core.info(
+            "No NPM_TOKEN found and no existing auth token - assuming trusted publishing or npm is already authenticated"
           );
         }
-      } else {
-        core.info("No user .npmrc file found, creating one");
+      } else if (npmToken !== undefined) {
+        core.info("No user .npmrc file found, creating one with NPM_TOKEN");
```
Andarist:
```suggestion
        core.info("No user .npmrc file found, creating one with NPM_TOKEN used as auth token");
```

akinoccc:
any updates?

bkeepers:
This breaks for anyone that was following the previous advice of setting `NPM_TOKEN=""`. Maybe update this to check if the token is not empty?

Andarist:
@bkeepers could you share the logs? I don't see how this would break things 🤔

Issue #523, opened 2025-09-16
GitHub Actions doesn’t trigger itself + non-executable files and directories


daniel-richter:
**What's going on?**
The changesets PR contains commits through the [changeset/action](https://github.com/changesets/action/) that seems to not trigger other GitHub Actions workflows.

**Why is this happening?**

1. **GitHub Actions doesn’t trigger itself.**

   The `changeset/action` by default acts on behalf of the _github-actions_ user:

   <img width="865" height="69" alt="Image" src="https://github.com/user-attachments/assets/81032b2e-7285-4047-81d9-e449b7addc50" />

   According to the [GitHub Actions "Triggering a workflow" docs](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow):

   > When you use the repository's GITHUB_TOKEN to perform tasks, events triggered by the GITHUB_TOKEN, with the exception of workflow_dispatch and repository_dispatch, will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs. **For example, if a workflow run pushes code using the repository's GITHUB_TOKEN, a new workflow will not run even when the repository contains a workflow configured to run when push events occur**. For more information, see [Use GITHUB_TOKEN for authentication in workflows](https://docs.github.com/en/actions/security-guides/automatic-token-authentication).

   The proposed solution is:
   > If you do want to trigger a workflow from within a workflow run, you can use a GitHub App installation access token or a personal access token instead of GITHUB_TOKEN to trigger events that require a token.

2. **Act as as different user.**

   Even in case the `changeset/action` gets an access token of another user as parameter, the action still acts on behalf of _github-actions[bot]_. 

   ```diff
    - uses: changesets/action@v1
      with:
        ...
   +   env:
   +     GITHUB_TOKEN: ${{ secrets.DEVOPS_SERVICEUSER_PAT }}
   ```

   According to the [Changesets Release Action docs](https://github.com/changesets/action?tab=readme-ov-file#inputs), the is another parameter `commitMode`:
   > commitMode - Specifies the commit mode. Use `git-cli` to push changes using the Git CLI, or `github-api` to push changes via the GitHub API. **When using `github-api`, all commits and tags are GPG-signed and attributed to the user or app who owns the `GITHUB_TOKEN`**. Default to `git-cli`.

   ```diff
    - uses: changesets/action@v1
      with:
        ...
   +     commitMode: github-api  
      env:
[...]

PR #563, opened 2026-01-27
Don't error on symlinks and executables when they don't have to be touched anyway

Andarist:
fixes https://github.com/changesets/action/issues/523

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "@actions/core": "^1.11.1",
     "@actions/exec": "^1.1.1",
     "@actions/github": "^6.0.1",
-    "@changesets/ghcommit": "2.0.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -40,7 +40,7 @@
     "@actions/core": "^1.11.1",
     "@actions/exec": "^1.1.1",
     "@actions/github": "^6.0.1",
-    "@changesets/ghcommit": "^2.0.0",
+    "@changesets/ghcommit": "2.0.1",
```
bluwy:
```suggestion
    "@changesets/ghcommit": "^2.0.1",
```

Maybe we should keep the caret

bluwy (approved):

Issue #525, opened 2025-09-29
Make `GITHUB_TOKEN` env var optional


bluwy:
> I'm currently working on this locally but thought to bring up as an issue first

Currently, the docs mention setting:

```yaml
env:
  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

when using the action. However, github actions can by default retrieve this using inputs, e.g. [actions/checkout](https://github.com/actions/checkout/blob/ff7abcd0c3c05ccf6adc123a8cd1fd4fb30fb493/action.yml#L12-L24) and [actions/setup-node](https://github.com/actions/setup-node/blob/89d709d423dc495668cd762a18dd4a070611be3f/action.yml#L21-L23).

It would be nice if `changesets/action` can do the same so users don't have to specify the env var manually. 

---

Note: Since `@changesets/get-github-info` still relies on the `GITHUB_TOKEN` env var, the action could internally set it from the input if it's unset.

PR #564, opened 2026-01-27
Automatically use the GitHub-provided token to allow most users to avoid explicit `GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}` configuration

Andarist:
closes https://github.com/changesets/action/issues/525

It feels a little bit off to even be able to do this implicitly but I don't see any scenarios in which one wouldn't like to pass this token to Changesets so I guess it's fine.

### DIFF
--- a/.changeset/quiet-vans-fail.md
+++ b/.changeset/quiet-vans-fail.md
@@ -1,0 +1,5 @@
+---
+"@changesets/action": minor
+---
+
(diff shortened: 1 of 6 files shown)


Chavrzp28 (commented):
Espero ligar codigo para más funciones de detección de errores críticos

PR #613, opened 2026-05-07
Alternative preview release flow

emmatown:
Example run: https://github.com/emmatown/changesets-action/actions/runs/25488677177

Note I've done this from a fork of this repo, not on changesets/action itself, this is partly because GitHub won't let you trigger the workflow dispatch unless the workflow exists on main and partly to demonstrate how it works on a fork.


.github/workflows/preview-release.yml
Andarist, line 1: nit, the "base" workflow is caled publish - perhaps we should rename this one to match the base one more closely?

scripts/release-pr.ts
```diff
@@ -4,7 +4,7 @@ import path from "node:path";
 import pkgJson from "../package.json" with { type: "json" };
 
 const tag = `v${pkgJson.version}`;
-const releaseLine = "pr-release";
+const releaseLine = `built-${process.env.GITHUB_REF_NAME}`;
```
Andarist, line 7: git stores branches as directories, perhaps we should use a pattern like this:

```suggestion
const releaseLine = `built-pr/${process.env.GITHUB_REF_NAME}`;
```

this way they would be "grouped"

.github/workflows/preview-release.yml
```diff
@@ -0,0 +1,49 @@
...
+permissions: {}
+
+jobs:
+  release:
+    timeout-minutes: 20
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
Andarist, line 15: how strong guarantees do we have here when it comes to what commit gets actually built and released? a workflow could get queued and also retried later - I feel this might be risky-ish because, I'd assume, it might run against the "current" version of a branch

a security researcher emailed me with a description of an atack like this:
1. the maintainer schedules a job on a PR
2. the attacker polls the github API against the triggering action (like a PR comment, or maybe a new workflow run)
3. the attacker quickly pushes malicious stuff to that PR

It feels like this workflow would be prone to this attack. I know none of those preview builds can truly be trusted - but we should minimize the risks when we can.

Andarist (commented):
With this approach - I wouldnt be able to trigger the preview release on a fork, right?

Andarist:
I think this is quite easier to use: https://github.com/changesets/action/pull/628

.github/workflows/preview-release.yml
kingsarhan: I tried to contact you through email for help/ work

PR #618, opened 2026-05-15
Migrate to oxfmt

bluwy:
Enables import sorting and `trailingComma: "always"` so matches the changesets repo setup.

NOTE: It also sorts the package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,40 +1,20 @@
 {
   "name": "@changesets/action",
   "version": "1.8.0",
-  "main": "dist/index.js",
-  "type": "module",
   "license": "MIT",
-  "devDependencies": {
-    "@changesets/changelog-github": "^0.5.1",
-    "@changesets/cli": "^2.29.3",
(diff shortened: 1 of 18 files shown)


package.json
```diff
@@ -53,12 +33,27 @@
     "semver": "^7.5.3",
     "unified": "^8.3.2"
   },
-  "pnpm": {
-    "overrides": {
-      "trim": "^0.0.3",
-      "y18n": "^4.0.1"
-    }
```
bluwy, line 60: This is completely removed as pnpm v11 no longer reads it, only works when configured in pnpm-workspace.yaml.

It doesn't seem like we configure that and everything's working, so I didn't port it to pnpm-workspace.yaml.

Andarist (dismissed):

Andarist:
@bluwy this requires conflict resolution now

Andarist (approved):

PR #619, opened 2026-05-15
Update vitest to v4

bluwy:
This will use rolldown and then later I could look into refactoring the build script to use rolldown


Andarist (approved):

Andarist:
> and then later I could look into refactoring the build script to use rolldown

It would be good to compare the build artifact size before and after.

PR #621, opened 2026-05-15
Use rolldown for bundling

bluwy:


pnpm-workspace.yaml
```diff
@@ -1,8 +1,11 @@
...
-
 blockExoticSubdeps: true
 dedupePeers: true
 dedupePeerDependents: true
 minimumReleaseAge: 10080
+
+overrides:
+  lightningcss: '-' # we do not bundle any css
+  postcss: '-' # we do not bundle any css
+  rolldown: $rolldown # use same version for vitest's rolldown
```
bluwy: Skip installing lightningcss and postcss as we don't have ay css files. Saves about 11MB to install.

Also set a consistent rolldown version here to prevent duplicate because vite is gated to an older version because of minimumReleaseAge 7 days :(

Andarist: I installed the new ones with some help from `minimumReleaseAgeExclude`

bluwy: I think I didn't went with it initially because we'd have to constantly update or maintain the versions, but I'm fine with anything for now.

Andarist:
### before 
```sh 
wc -c dist/index.js # 1195129 
gzip -c dist/index.js | wc -c # 368726
```

### after 
```sh
wc -c dist/index.js # 1228159 
gzip -c dist/index.js | wc -c # 359640 
```

Interestingly, this is a 3~% regression for the size of the bundle BUT a 3~% improvement for a gzipped one 😅

pnpm-workspace.yaml
Andarist: True, but at least we have a longer cooldown period and we still can explicitly opt into trusting certain packages/versions. We just should periodically prune this list. I think the tradeoff is fine.

Andarist (approved):

bluwy:
Yeah I think rolldown/oxc still have some work to get on par with terser for minifying, but it's not too much and we get much simpler tooling

Issue #1, opened 2019-09-16
Comment on issues and PRs that are fixed/included in releases to say something like "this fix for issue/this PR is included in x release"


mitchellhamilton:
Inspired by https://github.com/semantic-release/github

Not 100% sure about this because it could be spammy given that releases are batched in changesets but that problem doesn't really exist in semantic-release given that releases aren't batched

enesozturk:
Hey @mitchellhamilton, this looks pretty old but do you know this is developed?

PR #625, opened 2026-05-18
Comment changeset status in PRs

bluwy:
You can test the action with `uses: changesets/action/comment-pr-changeset@comment-pr-changeset-dist`.

<details>
<summary>Example workflow setup</summary>

```yaml
# .github/workflows/comment-changeset-pr-status.yml
name: Comment changeset status in PRs

on:
  pull_request_target:
    types: [opened, synchronize, reopened]

concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
  cancel-in-progress: true

jobs:
  pr-status:
    runs-on: ubuntu-slim
    permissions:
      contents: read # to check out files in the repo
    outputs:
      commentBody: ${{ steps.pr-status.outputs.commentBody }}
    steps:
      - name: Check out repo
        uses: actions/checkout@v6

      - name: Generate status
        id: pr-status
        uses: changesets/action/pr-status@v1

  comment:
    needs: pr-status
    runs-on: ubuntu-slim
    permissions:
      pull-requests: write # to create and update comments on PRs
    steps:
      - name: Comment on PR
        uses: mshick/add-pr-comment@v3
        with:
          message-id: changeset-pr-status
          message: ${{ needs.pr-status.outputs.commentBody }}
```

</details>

See this action ... in action: https://github.com/bluwy/changesets-playground/pull/1 (feel free to send PRs from forks for testing)

This actions focuses only on **generating the comment**, and defer the commenting to other actions. This is because we want to use reduced permissions for generating the comment to prevent risks of executing untrusted code and accessing the elevated github token. When changesets generate a release plan, we cannot guarantee in the future it will not execute code because of hooks/plugins or such.

---

When you have this workflow in your repo, you should disable the changeset bot for the repo so messages don't get reported twice. For this action, you should get the exact behaviour and messages from a bot. I didn't change the message for now.


src/pr-status/message.ts
Andarist, line 1: q: are the non-index files here 1to1 with what we have in the bot? or are there some significant differences between both?

rolldown.config.js
```diff
@@ -1,13 +1,16 @@
-import { defineConfig } from 'rolldown'
+import { defineConfig } from "rolldown";
 
 export default defineConfig({
-  input: 'src/index.ts',
+  input: {
+    index: "src/index.ts",
+    ["comment-pr-changeset"]: "src/comment-pr-changeset/index.ts",
```
Andarist: we have `changeset status` that strongly corresponds to those comments, right? what do you think about naming this `pr-status-comment`?

src/comment-pr-changeset/index.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  core.info("Creating comment message...");
+  const commentBody = await getCommentMessage(context);
+  const commentParam: CreateCommentParams | UpdateCommentParams = {
+    repo: context.base.repo.name,
+    owner: context.base.repo.owner.login,
+    issue_number: context.number,
+    body: commentBody,
+  };
+  core.setOutput("commentBody", commentBody);
```
Andarist: I'd probably only expose this if requested and when well-motivated

src/pr-status-comment/index.ts
```diff
@@ -0,0 +1,64 @@
...
+  Parameters<Octokit["rest"]["issues"]["createComment"]>[0]
+>;
+type UpdateCommentParams = NonNullable<
+  Parameters<Octokit["rest"]["issues"]["updateComment"]>[0]
+>;
+
+(async () => {
+  const context = github.context.payload.pull_request;
+  if (!context) {
+    core.error("This action should only be run on pull_request_target events");
```
Andarist: this is scary 😬

comment-pr-changeset/README.md
```diff
@@ -0,0 +1,47 @@
...
+        run: |
+          git fetch "$REPO" "$REF"
+          git switch -c pr "$SHA"
+        env:
+          REPO: ${{ github.event.pull_request.head.repo.clone_url }}
+          REF: ${{ github.event.pull_request.head.ref }}
+          SHA: ${{ github.event.pull_request.head.sha }}
+
+      - name: Comment changeset status
+        uses: changesets/action/comment-pr-changeset@v2
```
Andarist: `v2` doesn't exist yet - do we want to wait with merging this PR until right before the CLI v3 release (combined with `changesets/action` v2 release)?

pr-status-comment/README.md
```diff
@@ -0,0 +1,47 @@
+# changesets/action/comment-pr-changeset
+
+This action comments on PRs of its changeset status, e.g. whether it has changeset files and which packages will be released if the PR is merged.
+
+The action requires the base ref (of the repo) and head ref (of the PR) to be checked out in order to infer the changed files and packages. It also requires the [`pull_request_target`](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target) event to be triggered in order to have permissions to comment on the PR and to work in PRs from forks.
```
Andarist: What do you think about using the REST API (like the bot) to compute the diffs and changed packages? That would allow us to avoid relying on `pull_request_target`

comment-pr-changeset/README.md
```diff
@@ -0,0 +1,47 @@
+# changesets/action/comment-pr-changeset
```
mihkeleidast: Just a thought, but given https://github.com/changesets/check-action used to exist for this exact purpose before the bot, what do you think about reviving that repo & moving this action there, instead of keeping everything in one repo? Separate versioning might be useful.

Otherwise we will have an archived repo that directs users to the bot, and then presumably in the bot repo we'll have a note that directs users to another action repo 🙃

src/pr-status/message.ts
bluwy: For the messages and templates, they are the same except some small details:
- For `### 🦋 Changeset detected` and `### ⚠️ No Changeset found` they were two spaces around the emojis before. I made it a single space.
- I inject a `<!-- changesets-action-comment-pr-changeset -->` marker in the comment to update it later.

The rest of the wording is the same.

rolldown.config.js
bluwy: I named it this way to anticipate https://github.com/changesets/action/issues/1 being implemented one day. It depends if you'd consider that feature to be a "pr status comment" too? I'm fine with any naming though.

src/comment-pr-changeset/index.ts
bluwy: I'll remove it then, together with `commentId`?

pr-status-comment/README.md
bluwy: I figured the tradeoffs initially, but the REST API feels very fragile overall and required re-implementing a lot of the behaviour from the core packages.

But we can't get away with `pull_request_target` either with that. We need it to get permissions to comment on PR which you can never get in `pull_request` without a custom PAT. So at the end I find it better to leverage git and re-use the APIs we already built from core.

comment-pr-changeset/README.md
bluwy: Ah actually I meant v1 here. It would be nice to release this early if possible so we can get some user-testing, and we can decide to change anything for v2.

comment-pr-changeset/README.md
Andarist: > Otherwise we will have an archived repo that directs users to the bot, and then presumably in the bot repo we'll have a note that directs users to another action repo 🙃

😂 this might have to be cleaned up

> Just a thought, but given https://github.com/changesets/check-action used to exist for this exact purpose before the bot, what do you think about reviving that repo & moving this action there, instead of keeping everything in one repo? Separate versioning might be useful.

Likely, we'll add a couple of other subactions here soon and they will share some of the implementation with the main entry. So it feels it's just going to be easier to manage everything from this single repo.

bluwy: I prefer keeping it here since I'm thinking we could have more sub-actions in the future to manage changesets. For example, to split version and publishing we can also have sub-actions for them too.

I checked the github traffic for `check-action` and we only got 6 visits in 14 days, so I think it's ok to set that aside for now.

rolldown.config.js
bluwy: I'll rename it now and we can decide again what is better.

comment-pr-changeset/README.md
```diff
@@ -0,0 +1,47 @@
...
+        uses: actions/checkout@v6
+
+      - name: Check out head ref
+        run: |
+          git fetch "$REPO" "$REF"
+          git switch -c pr "$SHA"
+        env:
+          REPO: ${{ github.event.pull_request.head.repo.clone_url }}
+          REF: ${{ github.event.pull_request.head.ref }}
+          SHA: ${{ github.event.pull_request.head.sha }}
```
Andarist: Hm, it would be nice to just hide this in the action itself. It should be possible to just fetch the `github.event.pull_request.head.sha` and perhaps even avoid switching to that branch altogether. Just fetch the objects and diff (maybe we'd have to add some options to our git utils in the core to actually pull this off)

src/comment-pr-changeset/index.ts
Andarist: commentId is simple enough and could be useful, so I'd keep it

comment-pr-changeset/README.md
bluwy: Yeah it's possible to checkout in a detached state so we don't need a name, but I wasn't sure how well it'll work. (Probably fine?)

I figured it's better to be explicit here since this changes the state of the working directory, and every step has an explicit behaviour, so our action stays focused about comments only. And in case the user needs to modify how it's fetched.

src/comment-pr-changeset/index.ts
bluwy: Too late 😄 I've already logged the id in the logs, so considering the possible use-case without `commentBody` if we just have `commentId`, I think it's not a lot? I think we can also wait for it to be requested.

comment-pr-changeset/README.md
bluwy: I've updated the docs to v1

comment-pr-changeset/README.md
Andarist: I was thinking about making it so that it doesn't even change the state of the working directory. It should be possible to run the `git diff` against just two commits like `git diff current_sha fetched_sha` (or similar). That way, everything would be hidden in the action and it would be less likely that people accidentally run install scripts or something after they check out the PR's code

bluwy: Hmm theoretically it could work, but it'd need a lot of our APIs to be modeled around it or we have to re-implement a lot still. Eg getPackages need to work with the diff too if a PR adds a new package, or if the PR enters pre mode/updates the config, etc. The surface is a lot to maintain vs documenting to be careful.

Andarist: I vibed two experiments (I have not verified they work, I have vibed the implementation - but not the ideas):
- no PR checkout, use git objects directly to assemble the release plan: https://github.com/changesets/action/commit/6149ead367b6b932a9f79ad91aeb92d0ef8b4d89
- detached worktree checkout, this one automates the checkout, doesn't change the current local state of the filesystem and auto-deepens shallow clones: https://github.com/changesets/action/commit/55a0b3b9fc84aefd47b106304931a1300ead4106

I'd prefer to make this stuff as easy (and as mistake-free) as possible. What do you think about both of those approaches?

bluwy: If I had to choose between two, the second one would definitely be better as we can work on the filesystem directly. The shallow clone handling is nice too, I didn't handle that.

I don't know if it's too far, maybe we can handle `actions/checkout` as well? So we could clone in a temp dir and use normal branches instead of worktrees? 

Anyways, I'm happy with anything that gets us still working with the filesystem. Maybe we can move some of these git stuff to `@changesets/git` later. Do you want to send a PR of that code or commit directly?

Andarist: > I don't know if it's too far, maybe we can handle actions/checkout as well? S

There is no good or bad here - but the manual~ checkout is pretty normal and we already require it in the main action, so perhaps an explicit `actions/checkout` will just be symmetric~ and expected here.

> So we could clone in a temp dir and use normal branches instead of worktrees?

The worktree sounds to me a like a good fit here - cause it's nice not to juggle the local checkout and be able to use it as `cwd` and still have everything connected to the same git objects etc.

>  Maybe we can move some of these git stuff to @changesets/git later.

Yeah, that was my overall goal - that auto-deepening should be likely done before the `git merge-base` call we have there. That said, this code here falls out of a simple codepath because it needs to auto-deepen two remotes. So I'm not sure if it's going to be as easy to fit this into `@changesets/git` API assumptions. Worth a try though.

>  Do you want to send a PR of that code or commit directly?

As I mentioned, this was vibed :P so an explicit PR to your PR will be best. I have only skimmed it myself so far and not read through all of it properly

src/pr-status-comment/index.ts
bluwy: I've loosen up the message to also mention pull_request

comment-pr-changeset/README.md
bluwy: resolved in #626 (merged into this pr)

bluwy:
This is now ready again. The action now works slightly differently to prevent token leaks when generating the release plan. See the PR description and the added README for more info. I also renamed it as `changesets/action/pr-status` instead of `pr-status-comment` (maybe just `check` would be cool?)

I was also considering adding a simple comment action so we don't have to recommend a third-party action, but I don't want to widen the scope for now.

pr-status/README.md
```diff
@@ -0,0 +1,61 @@
...
+        uses: changesets/action/pr-status@v1
+
+  comment:
+    needs: pr-status
+    runs-on: ubuntu-slim
+    permissions:
+      pull-requests: write # to create and update comments on PRs
+    steps:
+      - name: Comment on PR
+        uses: mshick/add-pr-comment@v3
```
mihkeleidast: Hm. I like that this is split into two now security-wise, however... I was looking forward to using this new action at work in GHES, as we can't use the bot there. But we also have that nasty requirement that all third party actions must be from verified orgs on GitHub. Given this particular action is from an individual account, we definitely can't use it, so we'd still have to fork and own a copy.

Would be super nice if there was an alternative from a verified org, or if this was still provided as a changesets sub-action. Given all the supply chain security issues lately, I'd be much more comfortable using something directly from changesets and not bring in another variable.

bluwy: We could create a simple comment sub-action here but it'd be really simple and we can't commit to full-functionality like the other ones. IMO it's too strict to only limit to verified orgs, given pinning an action to a sha already saves you from attacks.

There's also https://github.com/marketplace/actions/create-or-update-comment by a github staff but setting up comment updates is slightly involved, but I guess that might not work for your setup too.

mihkeleidast: I totally agree that the verified org limitation is quite stupid and does not really protect against anything on it's own, but I don't make the rules :)

bluwy: I've created a simple action here https://github.com/changesets/action/pull/636

.changeset/wacky-taxis-sink.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/action": minor
+---
+
+Add a new `@changesets/action/pr-status` sub-action to generate the changeset status comment for PRs
```
Andarist:
```suggestion
Add a new `@changesets/action/pr-status` sub-action to generate the changeset status comment for PRs as an alternative to the [Changesets Bot](https://github.com/apps/changeset-bot).
```

pr-status/README.md
```diff
@@ -0,0 +1,63 @@
...
+See the [action metadata](action.yml) for details on the inputs and outputs.
+
+> [!WARNING]
+> **Do not run untrusted code** when using the `pull_request_target` event. The example below only checks out code and does not run any code from the PR. Read more about the `pull_request_target` event in the [GitHub documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target).
+
+## Example setup
+
+```yaml
+# .github/workflows/comment-changesets-pr-status.yml
+name: Comment changesets status in PRs
```
Andarist: I'm not like 100% sure which is preferred to be honest
```suggestion
name: Comment Changesets status in PRs
```

pr-status/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
```
Andarist: Isn't it synonymous to just this?
```suggestion
  pull_request_target:
```

pr-status/README.md
```diff
@@ -0,0 +1,63 @@
...
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  pr-status:
+    runs-on: ubuntu-slim
+    permissions:
+      contents: read # to check out files in the repo
+    outputs:
+      commentBody: ${{ steps.pr-status.outputs.commentBody }}
```
Andarist:
```suggestion
      comment-body: ${{ steps.pr-status.outputs.comment-body }}
```

pr-status/README.md
```diff
@@ -0,0 +1,63 @@
...
+        uses: changesets/action/pr-status@v1
+
+  comment:
+    needs: pr-status
+    runs-on: ubuntu-slim
+    permissions:
+      pull-requests: write # to create and update comments on PRs
+    steps:
+      - name: Comment on PR
+        uses: changesets/action/pr-comment@v1
```
Andarist, line 56: `/pr-comment` doesn't exist - have you maybe forgotten to commit it?

pr-status/README.md
```diff
@@ -0,0 +1,63 @@
...
+      commentBody: ${{ steps.pr-status.outputs.commentBody }}
+    steps:
+      - name: Check out repo
+        uses: actions/checkout@v6
+
+      - name: Generate status
+        id: pr-status
+        uses: changesets/action/pr-status@v1
+
+  comment:
```
Andarist: just to match the context and the used action name - but ofc this doesn't matter much
```suggestion
  pr-comment:
```

src/pr-status/index.ts
```diff
@@ -0,0 +1,21 @@
+import * as core from "@actions/core";
+import * as github from "@actions/github";
+import { getCommentMessage } from "./message.ts";
+
+(async () => {
```
Andarist: given we now target node24, we could likely just leverage TLA and drop the IIFE pattern

pr-status/README.md
bluwy: I implemented this at https://github.com/changesets/action/pull/636. I had it separated so it's easier to review, but likely we need to merge both before releasing.

pr-status/README.md
bluwy: You're right. Didn't notice that was the default.

src/pr-status/index.ts
bluwy: Updated to TLA and followed how the official actions handle error handling

src/pr-status/worktree.test.ts
```diff
@@ -0,0 +1,156 @@
...
+    await git(forkRepo, ["config", "user.email", "test@example.com"]);
+    await git(forkRepo, ["checkout", "-b", "feature"]);
+
+    // log all files in forkRepo with dot files
+    console.log(
+      "Files in forkRepo:\n",
+      await exec("ls", ["-Ra"], { nodeOptions: { cwd: forkRepo } }).then(
+        (res) => res.stdout.trim(),
+      ),
+    );
```
Andarist: do we want to keep this in the test?

Andarist (dismissed):

src/pr-status/worktree.test.ts
bluwy: Ah probably no. Didn't notice this before

Andarist (approved):

PR #626, opened 2026-05-20
Checkout PR in a detached worktree

Andarist:


bluwy:
- I've refactored to use `tinyexec` for the commands so we can easily port it to core in the future if we want to.
- I've also updated the tests to use fs-fixture so it's a bit cleaner.

PR #632, opened 2026-05-25
Simplify getChangelogEntry

bluwy:
I copied the function from https://github.com/changesets/changesets/blob/eaea8a45aca36f2c87adf888e3a73cd2c3a9ee39/packages/release-utils/src/utils.ts#L33

Which I've updated for next in https://github.com/changesets/changesets/pull/1962. This removes a ton of dependencies and slightly reduces our bundle size:

- Before: 1,227.77 kB 
- After: 1,131.49 kB  

We also don't have to update the deps (they're outdated). We could import this from `@changesets/release-utils` if we want, but the changes I made is in `next`. And I'm also not sure if we want to start using `@changesets/release-utils` here. So I went with the simple approach and copy the function to here.

I totally understand if we do not feel like doing this now. I feel that it's ok since we've added more tests for it, and it only affects the version PR description. But I'll leave it to you to see if we should do this now, or later.

### DIFF
--- a/.changeset/green-cougars-listen.md
+++ b/.changeset/green-cougars-listen.md
@@ -1,0 +1,5 @@
+---
+"@changesets/action": patch
+---
+
+Simplify internal implementation to get changelog entries for a package version
--- a/package.json
+++ b/package.json
@@ -26,12 +26,7 @@
     "@manypkg/get-packages": "^1.1.3",
     "@octokit/core": "^7.0.6",
     "@octokit/plugin-throttling": "^11.0.3",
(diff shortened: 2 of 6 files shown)


Andarist:
I'm not a fan of parsing this with regex 😬 but I guess, we can try... it gives me edge-case creeps though.

bluwy:
Yeah I'm happy to revert if we get any reports. Since it affects the PR description only, I think it's relatively safe to test if this works and prep the same changes in core.

Andarist (approved):

PR #635, opened 2026-05-26
chore: add node_id to PublishOptions type

openscript:
If `node_id` is not added to `PublishOptions` it will cause a type error here:

https://github.com/changesets/action/blob/031358f743b5a6199bd7a39bdc8b469280983df9/src/run.ts#L427

### DIFF
--- a/src/run.ts
+++ b/src/run.ts
@@ -60,6 +60,7 @@ const createRelease = async (
 
 type PublishOptions = {
   script: string;
+  node_id: string;
   githubToken: string;
   octokit: Octokit;
   createGithubReleases: boolean;


src/run.ts
```diff
@@ -60,6 +60,7 @@ const createRelease = async (
 
 type PublishOptions = {
   script: string;
+  node_id: string;
```
Andarist, line 63: You have mentioned the lack of this would result in a type error at the `octokit.graphql` call site. I don't see how those 2 things are correlated. I can also see the typecheck job passes on the main branch... so I'm not really sure what does this PR solve

openscript: Hmm.. when I run, `pnpm tsc` I get:

```
Lockfile is up to date, resolution step is skipped
Already up to date
Done in 573ms using pnpm v11.1.1
src/run.ts:484:34 - error TS2339: Property 'node_id' does not exist on type 'PullRequestData'.

484       pullRequestId: pullRequest.node_id,
                                     ~~~~~~~


Found 1 error in src/run.ts:484
```

Sorry for making noise, probably it's something wrong with my fork.

PR #636, opened 2026-05-28
Add simple PR comment sub-action

bluwy:
Separate but complements #625

### DIFF
--- a/pr-comment/action.yml
+++ b/pr-comment/action.yml
@@ -1,0 +1,22 @@
+name: Changesets - PR Comment
+description: A simple GitHub Action to comment on PRs
+runs:
+  using: node24
+  main: ../dist/pr-comment.js
+inputs:
+  github-token:
+    description: "The GitHub token to use for authentication. Defaults to the GitHub-provided token."
+    required: false
+    default: ${{ github.token }}
+  body:
+    description: "The comment body to post on the PR."
+    required: true
+  update-id:
+    description: "If provided, the action will update the comment that contains this id, or create a comment with this id to be updated later."
+    required: false
+outputs:
+  comment-id:
+    description: "The comment id of the comment that was created or updated."
+branding:
+  icon: message-circle
+  color: blue
--- a/pr-comment/action.yml
+++ b/pr-comment/action.yml
@@ -12,8 +12,11 @@ inputs:
     description: "The comment body to post on the PR."
     required: true
   update-id:
-    description: "If provided, the action will update the comment that contains this id, or create a comment with this id to be updated later."
+    description: >
+      By default, the action will create and update a comment with this id. Pass a different id to
+      create and update a new comment, or pass an empty string to disable updating comments.
     required: false
+    default: "changesets-action-pr-comment"
 outputs:
   comment-id:
     description: "The comment id of the comment that was created or updated."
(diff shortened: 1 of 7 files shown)


pr-comment/action.yml
```diff
@@ -0,0 +1,22 @@
...
+  main: ../dist/pr-comment.js
+inputs:
+  github-token:
+    description: "The GitHub token to use for authentication. Defaults to the GitHub-provided token."
+    required: false
+    default: ${{ github.token }}
+  body:
+    description: "The comment body to post on the PR."
+    required: true
+  update-id:
```
Andarist, line 14: should we maybe use a default value for this one? It feels like almost all users will like the auto-update mode to be enabled for them so it would be easier if `update-id` wouldn't be required to get that

src/pr-comment/index.ts
```diff
@@ -0,0 +1,70 @@
+import * as core from "@actions/core";
+import * as github from "@actions/github";
+
+type Octokit = ReturnType<typeof github.getOctokit>;
+type CreateCommentParams = NonNullable<
+  Parameters<Octokit["rest"]["issues"]["createComment"]>[0]
+>;
+type UpdateCommentParams = NonNullable<
+  Parameters<Octokit["rest"]["issues"]["updateComment"]>[0]
+>;
```
Andarist, line 10: octokit package layout... requires acquired taste 🫠

src/pr-comment/index.ts
```diff
@@ -0,0 +1,70 @@
...
+    throw new Error(
+      "This action should only be run on `pull_request_target` or `pull_request` events",
+    );
+  }
+
+  const githubToken = core.getInput("github-token", { required: true });
+  const body = core.getInput("body", { required: true });
+  const updateId = core.getInput("update-id", { required: false });
+
+  const commentMarker = updateId ? `<!-- ${updateId} -->` : null;
```
Andarist: it feels like we should prefix this with something, just in case

pr-comment/action.yml
bluwy: I'm thinking that this could be used for other things in the future, like https://github.com/changesets/action/issues/1, so I didn't want the action to be built specifically for the `pr-status` action. Or if for some reason they use this action for other workflows that comment on PRs.

Maybe it's not worth planning for that now and we can break later if we want? What do you think?

src/pr-comment/index.ts
bluwy: Since other actions uses this trick too, I was thinking we allow a free-er marker so they can easily swap actions later. I guess probably also not worth planning for for now.
```suggestion
  const commentMarker = updateId ? `<!-- changesets-action-pr-comment:${updateId} -->` : null;
```

pr-comment/action.yml
Andarist: I think it's better for us to make this simpler for people. We should support `update-id: null/false` to opt out from the default behavior

bluwy: I've tested and `update-id: ""` or `update-id: null` will work, which becomes an empty string in the script. `false` would become `"false"` and I choose to not specially handle a false string for now.

Andarist (approved):

PR #644, opened 2026-06-02
update note regarding `pull_request_target`

beeequeue:
- changes the note about `pull_request_target` to a CAUTION instead of a WARNING
- merges the two sections about it into one
- adds a note about using the bot as an alternative

### DIFF
--- a/pr-status/README.md
+++ b/pr-status/README.md
@@ -11,7 +11,7 @@ See the [action metadata](action.yml) for details on the inputs and outputs.
 >
 > **Do not run untrusted code** when using the `pull_request_target` event.
 >
-> The example below only ***checks out*** and ***reads*** code, and does not ***execute*** any code from the fork.
+> The example below only **_checks out_** and **_reads_** code, and does not **_execute_** any code from the fork.
 >
 > Read more about the `pull_request_target` event in the [GitHub documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target).
 
--- a/pr-status/README.md
+++ b/pr-status/README.md
@@ -2,7 +2,16 @@
 
 This action generates the changesets status in PRs, e.g. whether it has changeset files and which packages will be released if the PR is merged.
 
-It requires the repo to be checked out, and automatically fetches the PR head ref into a temporary detached worktree in order to infer the changed files and packages. It also requires the [`pull_request_target`](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target) event to be triggered in order to have permissions to comment on the PR and to work in PRs from forks.
+It requires the repo to be checked out, and automatically fetches the PR head ref into a temporary detached worktree in order to infer the changed files and packages.
+
+> [!CAUTION]
+> **This action uses `pull_request_target` by default to support PRs from forks.**
+> 
+> Generally, **do not execute any code except for GitHub Actions** when using the `pull_request_target` event.
+> 
+> The example below only _checks out_ and does not _run_ any code from the PR.
+> 
+> Read more about the `pull_request_target` event in the [GitHub documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target).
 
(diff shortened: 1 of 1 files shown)


pr-status/README.md
```diff
@@ -2,7 +2,18 @@
...
 
-It requires the repo to be checked out, and automatically fetches the PR head ref into a temporary detached worktree in order to infer the changed files and packages. It also requires the [`pull_request_target`](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target) event to be triggered in order to have permissions to comment on the PR and to work in PRs from forks.
+It requires the repo to be checked out, and automatically fetches the PR head ref into a temporary detached worktree in order to infer the changed files and packages.
+
+See the [action metadata](action.yml) for details on the inputs and outputs.
+
+> [!CAUTION]
+> **This action uses `pull_request_target` by default to support PRs from forks.**
+>
+> Generally, **do not execute any code except for GitHub Actions** when using the `pull_request_target` event.
```
beeequeue: i'm still not entirely sure about the wording in this line, but i think it's a good guideline to give to people who don't know _anything_ about the subject

i considered adding a line saying "e.g. installing dependencies can lead to cache poisoning in the main repo", but decided against it

beeequeue: maybe it should say "do not execute any code except code inside GitHub Actions"?

pr-status/README.md
```diff
@@ -27,6 +35,8 @@ name: Comment Changesets status in PRs
 on:
   pull_request_target:
 
+permissions: {} # require explicitly stating all permissions in each job
```
beeequeue, line 38: this should override whatever default permissions the repo is configured to use, so the comment step can't accidentally get `content: write`

Andarist (dismissed):

bluwy (changes requested):
I'd like to suggest some changes but I don't have time now, blocking it for a while

pr-status/README.md
bluwy: I think previously the "do not run untrusted code" would fit better, since github actions can't guarantee that they won't run untrusted code either. Maybe we can give some examples of what not to do, but I don't think is necessary.

pr-status/README.md
```diff
@@ -13,10 +24,7 @@ jobs:
     # ...
 ```
 
-See the [action metadata](action.yml) for details on the inputs and outputs.
-
-> [!WARNING]
-> **Do not run untrusted code** when using the `pull_request_target` event. The example below only checks out code and does not run any code from the PR. Read more about the `pull_request_target` event in the [GitHub documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target).
+You can also use the [Changesets Bot](https://github.com/apps/changeset-bot) if you don't want an extra action in your repo, or are worried about the `pull_request_target` event.
```
bluwy, line 27:
```suggestion
You can also use the [Changesets Bot](https://github.com/apps/changeset-bot) if you don't want an extra action in your repo. Or use the `pull_request` event if you're worried of possibly executing untrusted code in the future.
```

I think if they're worried of the _target event, we can suggest without, it'll still work. So the only reason to keep using the bot is if they don't want to maintain a workflow at all

bluwy: I have some vague memory about writing docs about pull_request before. Maybe I had deleted it in the pr-status PR. Well if we want to reuse it.

beeequeue: I think getting the comment on fork PRs is one of the bigger benefits of it, since new contributors are more likely to forget or not know about Changesets.

So personally I would rather just the bot than a non-target workflow

bluwy: That's fair. Let's leave it as is then

pr-status/README.md
kingsarhan: Thanks for the help am seriously lost and can not get help they succeeded in cutting me from any help anything is appreciated

kingsarhan (approved):

pr-status/README.md
kingsarhan: Do not run untrusted code I guess Is better

kingsarhan (changes requested):

bluwy (dismissed):

bluwy (approved):

PR #654, opened 2026-06-10
Fix auth setup for Yarn Berry

Andarist:
1. this will also require changes in `@changesets/cli` to actually call Yarn when publishing
2. this is not compatible with Yarn Classic
3. this whole detection of an existing token in an existing user-level config file is wonky... perhaps we should just not do this at all here? maybe we should create `changesets/action/setup-auth` for those that don't use trusted publishing? the current approach doesn't even respect configuration options like `npmrcAuthFile `

### DIFF
--- a/.changeset/smart-parks-argue.md
+++ b/.changeset/smart-parks-argue.md
@@ -1,0 +1,5 @@
+---
+"@changesets/action": minor
+---
+
+Internal auth setup for provided `NPM_TOKEN` environment variable when using Yarn gets now written to Yarn Berry's `~/.yarnrc.yml`.
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,6 +1,7 @@
 import fs from "node:fs/promises";
 import path from "node:path";
 import * as core from "@actions/core";
+import { getPackages } from "@manypkg/get-packages";
 import { Git } from "./git.ts";
 import { setupOctokit } from "./octokit.ts";
 import readChangesetState from "./readChangesetState.ts";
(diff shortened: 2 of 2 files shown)


.changeset/smart-parks-argue.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/action": minor
+---
+
+Internal auth setup for provided `NPM_TOKEN` environment variable when using Yarn gets now written to Yarn Berry's `~/.yarnrc.yml`.
```
beeequeue, line 5: is there a big benefit from doing this compared to just using env vars?

if anything, storing the value in a file if we might not need to sounds like a bad idea

Andarist: I agree and I was under the impression that you kinda can't get away from storing it in `.npmrc` file. But, according to a fresh agent run, npm actually supports overlaying its configs with environment variables, so... this should work:
```js
  let changesetPublishOutput = await getExecOutput(script, undefined, {
    cwd,
    ignoreReturnCode: true,
    env: {
      ...process.env,
      GITHUB_TOKEN: githubToken,
      ...(process.env.NPM_TOKEN && {
        "npm_config_//registry.npmjs.org/:_authToken": process.env.NPM_TOKEN,
      }),
    },
  });
```

I would have to test this out to ensure it works but this would be a great way of improving this (thanks for the pushback!).

That said:
- pnpm doesn't seem to support this. It might support it in older versions (pre v11) because it just delegated to the npm CLI for a bunch of this stuff, but v11 reimplements publishing and it doesn't seem to support this
- yarn classic supports `YARN_AUTH_TOKEN` and `NPM_AUTH_TOKEN ` (yes, it supports an npm-named variable that npm itself doesn't support), and it also supports the weird `npm_config_*` shape too
- yarn berry supports `YARN_NPM_AUTH_TOKEN`

beeequeue: okay, so for yarn we can use env vars, but for pnpm and npm we _need_ to write it to a file, i see.

pnpm still supports `.npmrc` for just auth, and it can also be put in `~/.config/pnpm/auth.ini` or (i think) `~/.config/pnpm/rc`

bluwy:
> 3\. this whole detection of an existing token in an existing user-level config file is wonky... perhaps we should just not do this at all here?

I think we should not do it at all. We have a lot of issues around the token handling, we should leave for `actions/setup-node` to handle this. A setup [like this](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#publish-to-npmjs-and-gpr-with-npm), passing NODE_AUTH_TOKEN to `changesets/action` would already work for npm and pnpm.

Yarn needs [like this](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#yarn2-configuration), like what you did in the PR with .yarnrc.yml. It's not great but we can document it. One of the super annoying things with yarn these days.

Plus, there's not a lot of reason to use tokens in CI these days with them having a short time limit.

---

But if you really prefer to have a better DX for yarn, then I'd prefer `changesets/action/setup-auth` instead

Andarist:
related: https://github.com/pnpm/pnpm/issues/12314#issuecomment-4675496836

bluwy:
That linked pnpm issue shouldn't affect us I think, since it's only for the project/repo-level npmrc file. `setup-node` and us do it outside of the repo. But good to keep in mind if we decide to refactor later.

beeequeue:
just a note: using `setup-node` does mean that it will download another node version which may not be necessary, as the runners come with certain versions built in (the ones specified by `using: node24`)

Andarist:
> using setup-node does mean that it will download another node version which may not be necessary

I think this only happens if you provide some input that tells it to use a specific node version

.changeset/smart-parks-argue.md
Andarist: Hah, pnpm added env-based support just recently: https://pnpm.io/blog/releases/11.6#file-free-registry-auth-via-environment-variables

I think [my tweet](https://x.com/AndaristRake/status/2065146078115123464) could have something to do with that 😅

PR #656, opened 2026-06-15
Add `select-mode`, `version`, `pack` and `publish` sub-actions

bluwy:
Adds a very simple implementation of `/select-mode`, `/version`, and `/publish` sub-actions that reuses functions from the root action. I renamed some of the inputs, copied most of the logic over. There's probably more cleanup possible but I don't want to stray too far for now.

I also did not write any readme/docs as ~~I'm lazy~~ maybe it changes after the publish-plan PRs.

I made this PR mainly so we can make updates on it easier later, but I'm also fine closing this as starting from scratch with the publish-plan stuff if that's easier.


version/action.yml
```diff
@@ -0,0 +1,43 @@
...
+  pr-title:
+    description: "The pull request title. Default to `Version Packages`"
+    required: false
+    default: "Version Packages"
+  pr-commit:
+    description: "The commit message. Default to `Version Packages`"
+    required: false
+    default: "Version Packages"
+  pr-branch:
+    description: "Sets the branch in which the action will run. Defaults to `github.ref_name`."
```
Andarist: Unrelated to this PR - but it seems to me this is a somewhat broken/weird input. It only kinda sets the base branch to `pr-branch` (so maybe this should be called `base-branch`?) but **without** actually even checking out this base. So dunno, maybe it kinda works OK for workflows that checkout another base beforehand but we don't make any guarantees here about the outcomes to be particularly coherent.

version/action.yml
```diff
@@ -0,0 +1,43 @@
...
+    required: false
+    default: ${{ github.token }}
+  script:
+    description: "The command to use to publish packages"
+    required: false
+  pr-title:
+    description: "The pull request title. Default to `Version Packages`"
+    required: false
+    default: "Version Packages"
+  pr-commit:
```
Andarist: I wonder if this shouldn't be `pr-commit-message`? also, this isn't 100% tighly coupled to the PR domain - so perhaps this should just be `commit-message`? I don't feel super strongly about it though

src/version/index.ts
```diff
@@ -0,0 +1,56 @@
...
+  core.setFailed((err as Error).message);
+}
+
+async function main() {
+  const githubToken = core.getInput("github-token", { required: true });
+  const script = core.getInput("script");
+  const prTitle = core.getInput("pr-title", { required: true });
+  const prCommit = core.getInput("pr-commit", { required: true });
+  const prBranch = core.getInput("pr-branch");
+  const prDraft = core.getInput("pr-draft") ?? undefined;
```
Andarist: This actually won't work as expected because `getInput` always returns a string:
https://github.com/actions/toolkit/blob/4b9afa4c890de9d1a84c75bb617ee4d96207823a/packages/core/src/core.ts#L151-L163

For that reason, in other places, we are using `getOptionalInput`

bluwy: Ah yeah I mean to use `||` here

version/action.yml
bluwy: Yeah let's do `commit-message`. I was thinking of that too, but it's kinda long, but `commit` is also kinda vague

version/action.yml
bluwy: Ah yeah it should be `base-branch`. I only read the top of the code and though this was the name of the branch for the PR.

bluwy: Actually yeah it is kinda weird. Maybe `pr-base-branch` would be better here? Because we're only using it for the PR's base

Andarist: Yeah, I think `pr-base-branch` would be better. It is kinda weird-ish that it's not like auto-synced with the actual local checkout and the user intention... but the use case of running a workflow from A (when we infer the pr base branch from the payload), and then targeting B with PRs (and maybe checking out B in the workflow too) is not really standard. So I don't think any changes to our default behavior should be made

publish/action.yml
```diff
@@ -0,0 +1,29 @@
...
+  main: ../dist/publish.js
+inputs:
+  github-token:
+    description: "The GitHub token to use for authentication. Defaults to the GitHub-provided token."
+    required: false
+    default: ${{ github.token }}
+  script:
+    description: "The command to use to publish packages"
+    required: false
+  packed-artifact-id:
```
Andarist: is this a good input name? maybe `pack-dir-artifact-id` would be better?

src/run.ts
```diff
@@ -81,17 +86,36 @@ type PublishResult =
...
-  let changesetPublishOutput = await getExecOutput(script, undefined, {
+  let changesetPublishOutput: ExecOutput;
+  const execOptions: ExecOptions = {
     cwd,
     ignoreReturnCode: true,
     env: { ...process.env, GITHUB_TOKEN: githubToken },
-  });
+  };
+
+  if (script) {
```
Andarist, line 103: Note how with a custom publish `script` we don't handle `fromPackDir` at all. My best idea to handle this would be to also add support for passing this argument~ through CLI args. That way the scripts could forward `env` to the underlying `changesets publish` kinda naturally - without messing with flags forwarding. Thoughts?

src/pack/index.ts
```diff
@@ -0,0 +1,79 @@
...
+    publishPlanPath: publishPlanArtifactId
+      ? await downloadPublishPlanArtifact(tmpDir, Number(publishPlanArtifactId))
+      : undefined,
+  });
+
+  const packedArtifact = await artifact.uploadArtifact(
+    `changeset-pack-${Date.now()}`,
+    await getFiles(outDir),
+    outDir,
+  );
```
bluwy: The default retention-days is 90, maybe this should be lower to around 30 days so it matches the limit for env approval timeout. Or we also allow passing this as an option.

src/run.ts
bluwy: I'd prefer not relying on envs as it can be sometimes hard to follow who set what at when. Maybe we can support some sort of interpolation? Like `script: pnpm format && pnpm changeset publish $action_args` or `... publish --from-pack-dir $action_from_pack_dir`. Just need to make sure to prevent potential script injections.

OR do we even need to allow specifying a script now? Since the sub-actions are explicit invocations now, if they need a pre-command or post-command, just do it in a pre-post-step. If they need a different version/publishing scheme, then don't call our action.

But I think what you have now is also fine and then we extend it later.

src/pack/index.ts
Andarist: I'm OK with 30 as our internal default and we can always make this configurable later - if there will actually be any demand for it

src/run.ts
Andarist: > OR do we even need to allow specifying a script now? Since the sub-actions are explicit invocations now, if they need a pre-command or post-command, just do it in a pre-post-step.

Yeah, I also wondered about that. Removing it would allow us to gather actual use cases for custom scripts before we re-commit to supporting them.

That said, it would break `changesets/action`'s own publishing pipeline - or at least, it would make it more cumbersome~. But perhaps it wouldn't be too bad if only we'd expose `changesets/action/github-release`

bluwy: For `changesets/action`, I think our publish script could be called directly like before without using the sub-action. For version though, I guess we need like a pre/post hook to make changes before we commit 🤔

publish/action.yml
bluwy: If we're to keep this, I think `pack-artifact-id` is a bit nicer.

Andarist: I'm OK with that too - I just thought about `pack-dir-artifact-id` as that would be symmetric~ with `--from-pack-dir`

src/run.ts
Andarist: For now, we can stick to using the old root action - and figure this one later.

publish/action.yml
bluwy: Oh, in that case let's go with `pack-dir-artifact-id` then so it's consistent. Maybe we want to revisit the same naming for the publish plan too

Andarist: > Maybe we want to revisit the same naming for the publish plan too

I thiiink that's already consistent enough - `--from-plan`/`publish-plan-artifact-id`. But I guess we could want to consistently use "publish plan" or just "plan" in both of those. I don't mind the current naming scheme but I would also be OK with adjusting it.

bluwy: Yeah I mean `--from-plan` could probably be `--from-publish-plan`, but it's kinda long. But, probably good to be explicit and it's not a common flag to manually use too

Andarist: Ok, I’m starting that rename here: https://github.com/changesets/changesets/pull/2108

Andarist (approved):

bluwy (commented):
LGTM

PR #661, opened 2026-06-23
Pass the publishing app token to the release step

Andarist:


.github/workflows/publish.yml
```diff
@@ -73,7 +73,6 @@ jobs:
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
         with:
           persist-credentials: false
```
Andarist, line 75: We have incorrectly switched this from `persist-credentials: true` to `persist-credentials: false` and that broke the CI pipeline

bluwy (approved):

PR #663, opened 2026-06-23
Fix the computed publish plan path passed internally to `changeset pack` by the `/pack` subaction

Andarist:


src/select-mode/index.ts
```diff
@@ -64,7 +64,9 @@ async function getMode(): Promise<ModeResult> {
   const cwd = process.cwd();
   const publishPlanPath = path.join(
     process.env.RUNNER_TEMP ?? (await fs.realpath(os.tmpdir())),
-    `changeset-publish-plan-${Date.now()}.json`,
+    `changeset-publish-plan-${Date.now()}`,
+    // we need a stable filename here (in a unique dirname) so the artifact download can find this cleanly
```
Andarist, line 68: relies on https://github.com/changesets/changesets/pull/2111

bluwy (approved):

PR #666, opened 2026-06-23
Fix the `/version` subaction to not crash on missing `pr-base-branch` input

Andarist:


bluwy (approved):
I prefer directly calling and using `|| undefined` instead on tiny wrappers but I'm not going to fight for it

Andarist:
I'm not a huge fan of tiny wrappers like this either - but it feels those will give us some consistency across the codebase and they will make it harder to miss the "need" to do the `|| undefined` normalization

PR #668, opened 2026-06-24
Rename to kebab-case for root action inputs and outputs

bluwy:
Could also rename some to match the new version/publish sub-action names, but I decided to make this change step by step.


src/index.ts
```diff
@@ -20,22 +20,22 @@ import { fileExists, getOptionalInput } from "./utils.ts";
   const cwd = process.cwd();
 
   const octokit = setupOctokit(githubToken);
-  const commitMode = getOptionalInput("commitMode") ?? "git-cli";
-  const prDraft = getOptionalInput("prDraft");
+  const commitMode = getOptionalInput("commit-mode") ?? "git-cli";
```
Andarist, line 33: should we maybe throw on the old names to help people notice the renames?

Andarist (dismissed):

src/index.ts
bluwy: Yeah I think it makes sense to implement something simple to error on old names. IIRC GitHub would only warn. I'll do this tomorrow.

action.yml
```diff
@@ -40,18 +40,18 @@ inputs:
...
     description: A boolean value to indicate whether a publishing is happened or not
-  publishedPackages:
+  published-packages:
     description: >
       A JSON array to present the published packages. The format is `[{"name": "@xx/xx", "version": "1.2.0"}, {"name": "@xx/xy", "version": "0.8.9"}]`
-  hasChangesets:
+  has-changesets:
     description: A boolean about whether there were changesets. Useful if you want to create your own publishing functionality.
-  pullRequestNumber:
+  pull-request-number:
```
Andarist, line 54: that's `pr-number` in the `/version`. perhaps we should match it here? 
https://github.com/changesets/action/blob/a35a66bd1ec69b007f5cef9646f34a13eb610a41/src/version/index.ts#L56C19-L56C28

note that it's also missing in the declared `outputs` of that subaction: https://github.com/changesets/action/blob/a35a66bd1ec69b007f5cef9646f34a13eb610a41/version/action.yml#L40

action.yml
```diff
@@ -21,15 +21,15 @@ inputs:
...
     description: Sets up the git user for commits as `"github-actions[bot]"`. Default to `true`
     required: false
     default: true
-  createGithubReleases:
+  create-github-releases:
     description: "A boolean value to indicate whether to create Github releases after `publish` or not"
     required: false
     default: true
-  commitMode:
+  commit-mode:
```
Andarist, line 32: @beeequeue noticed (rightfully so) this isn't exactly tightly-coupled to committing. But I'm not sure if we have the appetitete to rename this stuff right now

action.yml
Andarist, line 14: A little bit out of this PR's scope but I feel those should be renamed to `publish-script` and `version-script`

Andarist (dismissed):

action.yml
bluwy: Yeah we should align the names eventually. In this PR I'd like to focus on kebab-casing only

action.yml
bluwy: Yeah I can follow up with this later

trueberryless (approved):
Looks good to me 🚀

Andarist (approved):

beeequeue (approved):

PR #670, opened 2026-06-24
Avoid putting the github token in the global `.netrc` file

Andarist:


.changeset/git-cli-server-url.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/action": patch
+---
+
+Derive the Git server URL from the GitHub Actions context when configuring git CLI authentication to supper GitHub Enterprise Server setups.
```
beeequeue:
```suggestion
Derive the Git server URL from the GitHub Actions context when configuring git CLI authentication to support GitHub Enterprise Server setups.
```

Andarist: Oops, it seems I'm hungry 🍴

src/github.ts
```diff
@@ -46,17 +50,51 @@ const checkIfClean = async (options: GitOptions): Promise<boolean> => {
   return !stdout.length;
 };
 
-export class Git {
-  readonly octokit: Octokit | null;
+export class GitHub {
+  readonly #githubToken: string;
+  readonly octokit: Octokit;
   readonly cwd: string;
+  readonly commitMode: CommitMode;
```
beeequeue, line 57: should this just be called `mode` instead? it affects more than just commits

Andarist: I can rename it here but I wouldn't like to touch the public option in this PR - and those 2 are the same thing so keeping the name consistent makes it easier to reason about. That said, I agree it's not a particularly accurate name

beeequeue (approved):

PR #672, opened 2026-06-25
Always push tags with the github api

bluwy:
fix #669

### DIFF
--- a/src/github.ts
+++ b/src/github.ts
@@ -114,25 +114,16 @@ export class GitHub {
   }
 
   async pushTag(tag: string) {
-    if (this.commitMode === "github-api") {
-      return this.octokit.rest.git
-        .createRef({
-          ...context.repo,
-          ref: `refs/tags/${tag}`,
-          sha: context.sha,
-        })
-        .catch((err) => {
-          // Assuming tag was manually pushed in custom publish script
-          core.warning(`Failed to create tag ${tag}: ${err.message}`);
-        });
+    try {
+      await this.octokit.rest.git.createRef({
+        ...context.repo,
+        ref: `refs/tags/${tag}`,
+        sha: context.sha,
+      });
+    } catch (err) {
+      // Assuming tag was manually pushed in custom publish script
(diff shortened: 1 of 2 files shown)


src/github.ts
```diff
@@ -114,25 +114,16 @@ export class GitHub {
...
-          core.warning(`Failed to create tag ${tag}: ${err.message}`);
-        });
+    try {
+      await this.octokit.rest.git.createRef({
+        ...context.repo,
+        ref: `refs/tags/${tag}`,
+        sha: context.sha,
+      });
+    } catch (err) {
+      // Assuming tag was manually pushed in custom publish script
```
Andarist, line 124: Hm, it's pretty crucial for those tags to be created so we can create github releases properly. I know this is preexisting - but it would be great if we could only swallow specific errors here.

src/github.ts
```diff
@@ -114,25 +114,16 @@ export class GitHub {
...
+      core.warning(`Failed to create tag ${tag}: ${(err as Error).message}`);
     }
-    await exec("git", ["push", "origin", tag], {
```
Andarist, line 129: Hm, how sure are we this new flow will resolve the issue. The previous code wasn't exactly pushing all tags at once either, right?

src/github.ts
```diff
@@ -114,25 +114,16 @@ export class GitHub {
   }
 
   async pushTag(tag: string) {
```
Andarist, line 116: We should add a comment here explaining why we are not using the git CLI at all here

src/github.ts
bluwy: Hmm actually you might be right. I'm checking the timeline and we were pushing all git tags at once up until https://github.com/changesets/action/pull/391 which refactors to call them separately in parallel. After that's merged, we haven't got comments of it not working, so I guess it's fixed?

I was assuming since we push in parallel, GitHub might still detect it, and the REST API is able to workaround it. But maybe this isn't a problem now. I haven't personally tested this manually.

bluwy:
Looks like this was already fixed so I'm going to close this. But the suggested improvements can be done later.

PR #674, opened 2026-06-25
Remove GITHUB_TOKEN env var support

bluwy:
This aligns with the behaviour with the other sub-actions, but with an error if the old pattern is still used.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -3,16 +3,16 @@ import * as core from "@actions/core";
 import { GitHub } from "./github.ts";
 import readChangesetState from "./readChangesetState.ts";
 import { runPublish, runVersion } from "./run.ts";
-import { fileExists, getOptionalInput } from "./utils.ts";
+import { fileExists, getOptionalInput, getRequiredInput } from "./utils.ts";
 
 (async () => {
-  // to maintain compatibility with workflows created before github-token input was introduced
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -3,16 +3,16 @@ import * as core from "@actions/core";
...
 (async () => {
-  // to maintain compatibility with workflows created before github-token input was introduced
-  // it's important to prefer the explicitly set GITHUB_TOKEN over the default token coming from github.token
-  let githubToken = process.env.GITHUB_TOKEN || core.getInput("github-token");
-
-  if (!githubToken) {
-    core.setFailed("Please add the GITHUB_TOKEN to the changesets action");
-    return;
+  const githubToken = getRequiredInput("github-token");
+  if (process.env.GITHUB_TOKEN && process.env.GITHUB_TOKEN !== githubToken) {
```
Andarist, line 10: If we have this "they have to match if both are specified" rule - shouldn't we kinda also enforce that in other subactions that use `github-token`?

Andarist (approved):

src/index.ts
bluwy: I think this is more for compat to push migration, so eventually in the next major we can remove this and we don't have to mind about GITHUB_TOKEN anymore.

PR #680, opened 2026-06-29
Split git tag and github release options

### LINKED ISSUE #247: disable push tags

### LINKED ISSUE #547: `createGithubReleases: false` disables git tag pushes too which is undocumented
`createGithubReleases: false` disables git tag pushes too which is undocumented, and imo unexpected.

The docs say:

> - createGithubReleases - A boolean value to indicate whether to create Github releases after `publish` or not. Default to `true`

There are no mention of git tags.

When using `git-cli` mode, the tags are created locally. But they are not pushed. One would have to manually push them afterwards by checking the `outputs.published` value. But this feels like a bug and not an intended behaviour.

When using the `github-api` mode, the tags are not created locally, because they would be created via the API, which also does not happen for the above reason.

I would like to use the action to create the git tags, so that another action which has the `on: push: tags` trigger can run the deployment, which currently is only possible with the `git-cli` method if you do not want the action to publish releases too.

bluwy:
fix #247
fix #547 

NOTE: the above issues requested documentation improvements, but I have not done them here yet. I think probably it's easiest to consolidate all the docs update once we've finished reworking the actions & sub-actions.

### Possible alternatives

~~1. I wonder if it should be `push-git-tags` instead of `create-git-tags`, because we'd actually only do the pushing.~~ (renamed)
1. This is a breaking change because `create-github-releases: false` previously also disabled push tags, but now it'd push by default, unless set `push-git-tags: true` explicitly.
  
    We can kinda prevent this breaking change by making `push-git-tags` always follow the value of `create-github-releases`, unless `push-git-tags` was explicitly set, but I feel it's a bit hard to follow compared to the current behaviour, where both are independent options, just that `create-github-releases: true` will always force `push-git-tags: true` because inherently that's how github releases work.

    Or we support some sort of `create-github-releases: "tag-only"` but this also seems a bit weird.

### DIFF
--- a/.changeset/chubby-geese-sink.md
+++ b/.changeset/chubby-geese-sink.md
@@ -2,6 +2,6 @@
 "@changesets/action": major
 ---
 
-Add a new `create-git-tags` option that complements `create-github-releases` to control specifically if git tags should be created but not GitHub releases.
+Add a new `push-git-tags` option that complements `create-github-releases` to control specifically if git tags should be created but not GitHub releases.
 
-If `create-github-releases` was previously set to `false`, which also indirectly disabled git tag creation, git tags will now be created instead by default. If this is not desired, set `create-git-tags` to `false` explicitly.
+If `create-github-releases` was previously set to `false`, which also indirectly disabled git tag creation, git tags will now be created instead by default. If this is not desired, set `push-git-tags` to `false` explicitly.
--- a/.changeset/chubby-geese-sink.md
+++ b/.changeset/chubby-geese-sink.md
@@ -1,0 +1,7 @@
+---
+"@changesets/action": major
+---
+
+Add a new `create-git-tags` option that complements `create-github-releases` to control specifically if git tags should be created but not GitHub releases.
+
+If `create-github-releases` was previously set to `false`, which also indirectly disabled git tag creation, git tags will now be created instead by default. If this is not desired, set `create-git-tags` to `false` explicitly.
--- a/action.yml
+++ b/action.yml
@@ -29,6 +29,12 @@ inputs:
     description: "A boolean value to indicate whether to create Github releases after `publish` or not"
     required: false
     default: true
+  create-git-tags:
+    description: >
+      Whether to create git tags after publish. If `create-github-releases` is set to `true`,
+      this option will also always be `true`.
+    required: false
+    default: true
   commitMode:
     description: >
       An enum to specify the commit mode. Use "git-cli" to push changes using the Git CLI,
--- a/action.yml
+++ b/action.yml
@@ -29,7 +29,7 @@ inputs:
     description: "A boolean value to indicate whether to create Github releases after `publish` or not"
     required: false
     default: true
-  create-git-tags:
+  push-git-tags:
     description: >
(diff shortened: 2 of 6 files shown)


action.yml
```diff
@@ -29,6 +29,12 @@ inputs:
     description: "A boolean value to indicate whether to create Github releases after `publish` or not"
     required: false
     default: true
+  create-git-tags:
```
Andarist: The proposded alternative of `push-git-tags` sounds nice

.changeset/chubby-geese-sink.md
```diff
@@ -0,0 +1,7 @@
+---
+"@changesets/action": major
+---
+
+Add a new `create-git-tags` option that complements `create-github-releases` to control specifically if git tags should be created but not GitHub releases.
+
+If `create-github-releases` was previously set to `false`, which also indirectly disabled git tag creation, git tags will now be created instead by default. If this is not desired, set `create-git-tags` to `false` explicitly.
```
Andarist: is `create-git-tags: false, create-github-releases: true` a valid combination? I get that technically somebody could handle tags completely in their custom script but those tags are pretty crucial for the ability to create github releases.

Was the motivation behind the PR to actually *push* tags when `create-github-releases: false`? Perhaps we could just skip github releases when that's set to false and avoid a new option relate to tags? In that scenariom the tags would be auto-pushed at all times.

bluwy: > is `create-git-tags: false, create-github-releases: true` a valid combination? I get that technically somebody could handle tags completely in their custom script but those tags are pretty crucial for the ability to create github releases.

It's not "valid" but internally we'd always push git tags anyways if it's configured like that.

> Was the motivation behind the PR to actually _push_ tags when `create-github-releases: false`? Perhaps we could just skip github releases when that's set to false and avoid a new option relate to tags? In that scenariom the tags would be auto-pushed at all times.

Yeah, people wanted to push tags only and not create releases. There are also people who dont want to push tags at all though (#247), so we need to support it in some way. We _could_ ask them to pass `--no-git-tag`, but maybe they still want to create the tags to process them later? Or we ask them to do `--no-git-tag` and use `changeset tag` in a separate step? But then the results might not align if there's failed-to-publish packages.

I think it's probably simpler to support configuring push tags directly so they don't have to go through the process 🤔

trueberryless (approved):
Amazing. Changes look good to me, the logic seems fine. Description matches the behaviour.

Do we have any tests yet for the actions?

.changeset/chubby-geese-sink.md
Andarist: > It's not "valid" but internally we'd always push git tags anyways if it's configured like that.

Shouldn't we at least throw when that combination gets used? It feels bad to ignore `create-git-tags` like this - unless I'm misinterpreting what u mean with "we'd always push git tags anyways if it's configured like that."

bluwy: Aight I'll add an error if that combination happens.

bluwy:
Our tests currently is very loose. We should get to that after we're finished with most of the big changes I think.

Andarist (approved):

PR #686, opened 2026-07-01
Setup git user in the release script

Andarist:
I missed this in https://github.com/changesets/action/pull/679

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -62,8 +62,8 @@ jobs:
       - name: Build
         run: pnpm build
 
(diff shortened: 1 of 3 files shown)


bluwy (approved):

.github/workflows/publish.yml
```diff
@@ -106,8 +100,15 @@ jobs:
       - name: Build
         run: pnpm build
 
+      - uses: ./.github/actions/setup-bot-auth
```
Andarist, line 103: This now creates the git identity. It usually would be created in our root action and the custom publish script would then use that - but since the split publishing PR we lost that as the git identity setup is not implemented in `/publish` subaction (as it's not really needed there)

PR #688, opened 2026-07-01
Remove `setup-git-user` in favor of setting fallback only when identity is missing

Andarist:


bluwy:
I think we can try with this setup. Can we also remove the `setup-git-user` option too? Also needs a changeset

Andarist:
@bluwy could u take another look?

bluwy (approved):

PR #692, opened 2026-07-06
Use GitHub API for pushes by default and require explicit `github-token` when using custom tokens

Andarist:
This should fix auth headers when the action gets used with `actions/checkout`'s default `persist-credentials: true`

related to https://github.com/sveltejs/kit/pull/16248


bluwy (approved):

src/github.test.ts
```diff
@@ -0,0 +1,290 @@
...
+        "refs/heads/changeset-release/main",
+      ]),
+    ).toBe(await git(repository, ["rev-parse", "HEAD"]));
+    expect(remote.requests.length).toBeGreaterThan(0);
+    expect(
+      remote.requests.map((request) => request.headers.authorization),
+    ).toEqual(remote.requests.map(() => [getAuthorization(actionToken)]));
+  }, 15_000);
+
+  it("uses the push URL when it differs from the fetch URL", async () => {
```
beeequeue, line 153: do we need to test for this? is it a case we handle explicitly?

src/github.ts
```diff
@@ -50,47 +48,115 @@ const checkIfClean = async (options: GitOptions): Promise<boolean> => {
...
+
+    // GIT_CONFIG_COUNT/KEY_n/VALUE_n add command-scoped config. Preserve any
+    // existing entries and append ours. `http.extraHeader` is multi-valued, so
+    // merely adding our Authorization header would make Git send both tokens.
+    // An empty value resets the list; the following value adds only our token.
+    //
+    // In v1, `github-token` lived in ~/.netrc. When checkout had already
+    // supplied Authorization through an extraheader, that header took
+    // precedence and ~/.netrc was effectively a fallback. These entries
+    // intentionally make `github-token` win for pushes.
```
beeequeue: i think a full description of how we override the token and why we do it that way should be written somewhere else, the current doc comments are both duplicated and split up so they are hard to read.

src/test-utils.ts
```diff
@@ -0,0 +1,287 @@
...
+  headers: NodeJS.Dict<string[]>;
+};
+
+function recordRequest(request: IncomingMessage): RecordedRequest {
+  return {
+    method: request.method ?? "GET",
+    url: request.url ?? "/",
+    headers: request.headersDistinct,
+  };
+}
```
beeequeue: this function should be closer to where it's used

src/test-utils.ts
```diff
@@ -0,0 +1,287 @@
...
+          if (error) {
+            reject(error);
+            return;
+          }
+          resolve();
+        });
+      });
+    },
+  };
+}
```
beeequeue, line 254: this should probably be in its own file and split up a bit.
i would also prefer if we use [`srvx`](https://srvx.h3.dev/guide/handler) so we can use Request/Response instead of node's http server

src/github.test.ts
beeequeue, line 1: i think these test files are getting a bit overwhelming... my brain struggles massively to read them

beeequeue (dismissed):

src/test-utils.ts
bluwy: IMO node's `createServer` isn't bad and we shouldn't try to avoid it. Its API is not the most modern, but we don't need to risk running a server framework that might have its own issues, plus Request-Response-based servers also have their own limitations which don't exist in `createServer`.

beeequeue: i just prefer Requst/Response over node's api 😜
srvx is the one used by h3/nuxt (soon?), so i think it'd be fine for our use case in tests.

like i wrote before, i'm getting overwhelmed by the test setups we're creating so anything to simplify them would be a win in my eyes

src/github.test.ts
Andarist: we handle it here:
https://github.com/changesets/action/blob/99664fb28f52e18384ddd82b657971d806d4025f/src/github.ts#L109-L121

src/github.ts
Andarist: I added a short explainer on why we are doing this stuff at the top of this function but kept comments relevant to specific code lines close to them (while improving them a little bit). I feel it's easier to follow the steps performed here this way

src/test-utils.ts
Andarist: Hm, I get this is a little bit hard to follow - but at the same time it's not something we need to care about deeply as this is test-only glue code. What matters is in the test bodies (which admittedly are kinda long right now). I added some code comments here to make it easier to follow.

I think it would be good to split up this file in the future - but given the test files are now in the same directory as the source files and this is a "single-package" repo, I don't want to restructure this too much right now.

src/github.test.ts
Andarist: I moved some shared stuff into in-file utils. I hope this will help with the readability of those tests

PR #428, opened 2024-12-04
docs: Update documentation to reflect changesets publish expected environment variable

EduardoAC:
# Changelog

- Documentation updated to assist with the usage of the action in combination with `changesets publish`

fixes https://github.com/changesets/changesets/issues/942

## Additional context

I wanted to add some background to this PR to provide better context for the reviewers to assist with the approval. The environment variable is defined by how NodeJS is set up within the GitHub Action Setup-node, which will automatically create a `.npmrc` file using this as an environment variable.

Therefore, the action expects us to use this variable over others, while if you decide to create your own `.npmrc,` we will be in control. As we can see on [here](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#publish-to-npmjs-and-gpr-with-npm). Hence the documentation updated to align with the previous steps above to align with the setup-node specifications mitigating potential issues when developer use the code given in the Readme

PR #695, opened 2026-07-16
Remove npm auth handling in favor of docs

### LINKED ISSUE #58: project's .npmrc not respected

### LINKED ISSUE #132: Document of publish with `NPM_TOKEN` not accurately while use actions/setup-node
I have make a POC commit and action here:

- Action log: https://github.com/zthxxx/getrusage.js/runs/4728936828?check_suite_focus=true
- Commit: https://github.com/zthxxx/getrusage.js/commit/41dbf4d44b4d7e1395934947cabb45660a6a79da

1. it use `actions/setup-node`, with the `registry-url` params
2. list all the npmrc for debug
3. drop `NODE_AUTH_TOKEN` for `changesets/action`

There is no doubt, this commit will make action status from `success` to **`failure`**.

### Reason

As the annotation in screenshot, `actions/setup-node` with the `registry-url` params will create "user" config `.npmrc` with `_authToken=${NODE_AUTH_TOKEN}`,

but it at `~/work/_temp/.npmrc` not `~/.npmrc`, as well as the `userconfig` field point to `~/work/_temp/.npmrc`.

That's said in [GitHub Actions Docs - Publishing packages to the npm registry](https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages#publishing-packages-to-the-npm-registry).

So that `changesets/actions` create and setup `_authToken` will not work.

### Proposed

Maybe update the document about that case?

### LINKED ISSUE #147: [2.1.0] .npmrc detection depends on HOME process env
Not a big issue, but since latest release the `.npmrc` detection depends on [`process.env.HOME`](https://github.com/changesets/action/blob/1ed9bc24b7a56462c183eb815c8f4bdf0e2e5785/src/index.ts#L55). 

It broke our release (private repo with github private repositories). HOME was somehow changed in the previous step by our build scripts, thus `.npmrc` was wrongly recreated.

## Workaround

To help if someone got hit.

```
      - name: Create Release Pull Request or Publish to GPR
        id: changesets
        uses: changesets/action@v1.2.0
        with:
          publish: yarn release
        env:
          HOME: ${{ github.workspace }}
```

## Fix

I would gladly send a fix (or a doc), but I'm not sure how `process.env.HOME` and the new `cwd` param should behave. Otherwise as I read that `.npmrc` related code will be deprecated / removed, just a note is fine

### LINKED ISSUE #178: pnpm github 401 Unauthorized
I am trying to publish monorepo packages to a private Github registry using `changesets/action`. 

Whenever I push to `main`, I get the following error:

```
npm notice 📦  @inception-health/utils@0.0.1
npm notice === Tarball Contents === 
npm notice 1.3kB dist/index.js  
npm notice 960B  package.json   
npm notice 284B  dist/index.mjs 
npm notice 209B  dist/index.d.ts
npm notice === Tarball Details === 
npm notice name:          @inception-health/utils                 
npm notice version:       0.0.1                                   
npm notice package size:  1.2 kB                                  
npm notice unpacked size: 2.7 kB                                  
npm notice shasum:        12d6dc8e84110XXXXXX
npm notice integrity:     sha512-tGyxr7tYjFXh2[...]XXXXXXXXXX
npm notice total files:   4                                       
npm notice 
npm ERR! code E401
npm ERR! 401 Unauthorized - PUT https://npm.pkg.github.com/@inception-health%2futils - Your request could not be authenticated by the GitHub Packages service. Please ensure your access token is valid and has the appropriate scopes configured.
```

I have the following configuration:

**.npmrc**

```
@inception-health:registry=https://npm.pkg.github.com/
//npm.pkg.github.com/:_authToken=$GITHUB_TOKEN
```
_Note_: I also tried to dynamically add the `.npmrc` during the workflow build

```yaml
      - name: Creating .npmrc
        run: |
          cat << EOF > "$HOME/.npmrc"
            @inception-health:registry=https://npm.pkg.github.com/
            //npm.pkg.github.com/:_authToken=$GITHUB_TOKEN
          EOF
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

**Github Workflow**

```YAML
name: Changesets
on:
  push:
    branches:
      - main
env:
  CI: true
  PNPM_CACHE_FOLDER: .pnpm-store
jobs:
  version:
    timeout-minutes: 15
    runs-on: ubuntu-latest
    permissions: write-all
    steps:
      - name: checkout code repository
        uses: actions/checkout@v2
        with:
          fetch-depth: 0
      - name: setup node.js
        uses: actions/setup-node@v2
        with:
          node-version: 14
          registry-url: https://npm.pkg.github.com/
          scope: "@inception-health"
      - name: install pnpm
        run: npm i pnpm@latest -g
      - name: setup pnpm config
        run: pnpm config set store-dir $PNPM_CACHE_FOLDER
      - name: install dependencies
        run: pnpm install
      - name: create and publish versions
        uses: changesets/action@v1
        with:
          version: pnpm ci:version
          commit: "chore: update versions"
          title: "chore: update versions"
          publish: pnpm ci:publish
          setupGitUser: false
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

_Note_: I also tried to replace the `GITHUB_TOKEN` with a custom PAT. Same error.

**Package.json**

For all the packages I wanted to publish, I included the following config

```JSON
{
   ...
  "publishConfig": {
    "access": "restricted",
    "registry": "https://npm.pkg.github.com/inception-health/"
  }
}
```

**CI Action**

This is the `pnpm` script for `ci:publish` that are included in the root `package.json`.

```JSON
{
 ...
  "scripts": {
    "ci:publish": "pnpm build && pnpm recursive publish",
    "ci:version": "pnpm changeset version"
  },
 ...
}
```

I don't know what is missing? I appreciate any guidance.

**Version**
- `pnpm`: 7.0.0-rc.9
- `@changesets/cli`: "^2.22.0"

### LINKED ISSUE #262: question: secrets to file system

### LINKED ISSUE #336: Publish not working for internal Artifactory registry
In my GitHub Workflow I'm creating a .npmrc with connection info for our internal Artifactory registry and then executing the changesets action. It fails on running the `pnpm changesets publish` command with an "incorrect or missing password" and to run npm login. The documentation states that in the presence of a .npmrc, that user defined registry info is used instead. So why does it seem that its failing to use it and expecting an NPM_TOKEN? Here's an excerpt from the failed workflow:

```
No changesets found, attempting to publish any unpublished packages to npm
Found existing user .npmrc file
Didn't find existing auth token for the npm registry in the user .npmrc file, creating one
/home/runner/setup-pnpm/node_modules/.bin/pnpm run changeset:release 
```

### LINKED ISSUE #342: Error "No user .npmrc file found" when file actually exists
I'm getting a 403 when trying to release a private page to the private Github registry.
The project is a fresh turbo project created with `yarn dlx create-turbo@latest` and a dummy `foo` package to test the release works.

The logs says the `.npmrc` file doesn't exist, but it does:

```
auto-install-peers = true
@myorg:registry=https://npm.pkg.github.com
//npm.pkg.github.com/:_authToken=${GITHUB_TOKEN}
```

My release workflow includes:

```yml
  id: changesets
  uses: changesets/action@v1
  with:
    version: yarn run changeset:version # translated as changeset version && yarn install --mode=update-lockfile
    publish: yarn run release-packages # translated as turbo run build --filter='./packages/*' && changeset publish
  env:
    GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # Token automatically added by Github
```

Error output, note the `No user .npmrc file found, creating one` line.

```
Run changesets/action@v1
  with:
    version: yarn run changeset:version
    publish: yarn run release-packages
    setupGitUser: true
    createGithubReleases: true
  env:
    YARN_VERSION: 4.0.2
    GITHUB_TOKEN: ***
setting git user
/usr/bin/git config user.name "github-actions[bot]"
/usr/bin/git config user.email "github-actions[bot]@users.noreply.github.com"
setting GitHub credentials
No changesets found, attempting to publish any unpublished packages to npm
No user .npmrc file found, creating one
/usr/local/bin/yarn run release-packages
• Packages in scope: @myorg/foo, @repo/eslint-config, @repo/typescript-config, @repo/ui
• Running build in 4 packages
• Remote caching disabled

No tasks were executed as part of this run.

 Tasks:    0 successful, 0 total
Cached:    0 cached, 0 total
  Time:    131ms 

🦋  info npm info @myorg/foo
🦋  error Received an unknown error code: E403 for npm info "@myorg/foo"
🦋  error 403 Forbidden - GET https://npm.pkg.github.com/@myorg%2ffoo - Permission permission_denied: read_package
🦋  error In most cases, you or one of your dependencies are requesting
🦋  error a package version that is forbidden by your security policy, or
🦋  error on a server you do not have access to.
Error: Error: The process '/usr/local/bin/yarn' failed with exit code 1
Error: The process '/usr/local/bin/yarn' failed with exit code 1
```

I've tracked the error to the source code, it's coming from this conditional, on the else block:

https://github.com/changesets/action/blob/2bb9bcbd6bf4996a55ce459a630a0aa699457f59/src/index.ts#L58-L59

I had `changesets` working for another turborepo monorepo with the only notably difference that it was using `pnpm`.
What is going wrong? perhaps `process.env.HOME`, but why?

### LINKED ISSUE #363: Support private registry
I've encounter problem when using changesets/action when I'm using the private registry that is not npm.

Upon looking into it the problem seems to coming from the bellow line where it only check for npm registry name.

```
  // check based on https://github.com/npm/cli/blob/8f8f71e4dd5ee66b3b17888faad5a7bf6c657eed/test/lib/adduser.js#L103-L105
  return /^\s*\/\/registry\.npmjs\.org\/:[_-]authToken=/i.test(line);
});
```

However, beside npm registry, we also have github registry and custom domain gitlab registry.
```
// npm
//registry.npmjs.org/:_authToken

// github
//npm.pkg.github.com/:_authToken=TOKEN

// custom domain gitlab registry
//code.org.com/packages/npm/:_authToken=
```

Hence I'm making the change here to address for this problem.
https://github.com/changesets/action/pull/362

Thank you the team for your work on maintaining this tool!

### LINKED ISSUE #424: Publish in custom artifactory

bluwy:
Issues:
- closes #58
- closes #132
- closes #147
- closes #178 
- closes #262
- closes #336
- closes #342
- closes #363
- closes #424

PRs:
- closes #173
- closes #219
- closes #235 
- closes #309
- closes #362
- closes #425
- closes #428
- closes #529
- closes #654
- closes #658

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ jobs:
         uses: actions/setup-node@v6
         with:
           node-version: 26
-          registry-url: https://registry.npmjs.org/
+          registry-url: https://registry.npmjs.org/ # set up .npmrc file for authentication
 
       - name: Install Dependencies
         run: pnpm install --frozen-lockfile
@@ -143,7 +143,7 @@ jobs:
         uses: actions/setup-node@v6
         with:
           node-version: 26
-          registry-url: https://registry.npmjs.org/
+          registry-url: https://registry.npmjs.org/ # set up .npmrc file for authentication
 
       - name: Install Dependencies
         run: pnpm install --frozen-lockfile
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ jobs:
 
 #### With Publishing
 
-Before you can setup this action with publishing, you'll need to have an [npm token](https://docs.npmjs.com/creating-and-viewing-authentication-tokens) that can publish the packages in the repo you're setting up the action for and doesn't have 2FA on publish enabled ([2FA on auth can be enabled](https://docs.npmjs.com/about-two-factor-authentication)). You'll also need to [add it as a secret on your GitHub repo](https://help.github.com/en/articles/virtual-environments-for-github-actions#creating-and-using-secrets-encrypted-variables) with the name `NPM_TOKEN`. Once you've done that, you can create a file at `.github/workflows/release.yml` with the following content.
+Check the [npm authentication guide](./docs/set-up-npm-auth.md) to set up publishing to npm. After that, an example workflow with publishing may look like this:
 
 ```yml
 name: Release
@@ -99,6 +99,7 @@ jobs:
         uses: actions/setup-node@v6
         with:
           node-version: 26
+          registry-url: https://registry.npmjs.org/
 
       - name: Install Dependencies
         run: pnpm install --frozen-lockfile
@@ -110,31 +111,7 @@ jobs:
           # This expects you to have a script called release which does a build for your packages and calls changeset publish
           publish-script: pnpm release
         env:
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
-
-      - name: Send a Slack notification if a publish happens
-        if: steps.changesets.outputs.published == 'true'
-        # You can do something when a publish happens.
-        run: my-slack-bot send-notification --message "A new version of ${GITHUB_REPOSITORY} was published!"
-```
-
-By default the GitHub Action creates a `.npmrc` file with the following content:
-
-```txt
-//registry.npmjs.org/:_authToken=${process.env.NPM_TOKEN}
-```
-
-However, if a `.npmrc` file is found, the GitHub Action does not recreate the file. This is useful if you need to configure the `.npmrc` file on your own.
-For example, you can add a step before running the Changesets GitHub Action:
-
-```yml
-- name: Creating .npmrc
-  run: |
-    cat << EOF > "$HOME/.npmrc"
-      //registry.npmjs.org/:_authToken=$NPM_TOKEN
-    EOF
-  env:
-    NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
 ```
 
 #### Custom Publishing
@@ -166,6 +143,7 @@ jobs:
         uses: actions/setup-node@v6
         with:
           node-version: 26
+          registry-url: https://registry.npmjs.org/
 
       - name: Install Dependencies
         run: pnpm install --frozen-lockfile
@@ -178,6 +156,8 @@ jobs:
         if: steps.changesets.outputs.hasChangesets == 'false'
         # You can do something when a publish should happen.
         run: pnpm publish
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
 ```
 
 #### With version script
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ jobs:
         uses: actions/setup-node@v6
         with:
           node-version: 26
-          registry-url: https://registry.npmjs.org/ # set up .npmrc file for authentication
+          registry-url: https://registry.npmjs.org/ # makes the action set up npm authentication
 
       - name: Install Dependencies
         run: pnpm install --frozen-lockfile
@@ -143,7 +143,7 @@ jobs:
         uses: actions/setup-node@v6
         with:
           node-version: 26
-          registry-url: https://registry.npmjs.org/ # set up .npmrc file for authentication
+          registry-url: https://registry.npmjs.org/ # makes the action set up npm authentication
 
       - name: Install Dependencies
         run: pnpm install --frozen-lockfile
--- a/docs/set-up-npm-auth.md
+++ b/docs/set-up-npm-auth.md
@@ -6,13 +6,6 @@ This is a brief guide on setting up npm authentication in GitHub Actions. Most o
 
 It is recommended by npm to use [Trusted Publishing](https://docs.npmjs.com/trusted-publishers), or [Staged Publishing](https://docs.npmjs.com/staged-publishing), or both, to securely publish packages from CI.
 
-Token-based publishing (with [Granular Access Tokens](https://docs.npmjs.com/about-access-tokens#about-granular-access-tokens)) is no longer recommended, with many restrictions that make it difficult to use in CI workflows. For example:
-
-- They expire after a maximum of 90 days, which requires periodic manual token rotation.
-- 2FA-bypass tokens are [being deprecated](https://github.blog/changelog/2026-07-08-npm-install-time-security-and-gat-bypass2fa-deprecation/#2fa-bypass-tokens-will-no-longer-publish-directly) and will soon be not allowed to publish packages with 2FA enabled.
-
-However, if you're using a different npm-compatible registry that does not support Trusted Publishing or Staged Publishing, you may still opt for token-based publishing. Check out the [next section](#token-based-publishing) for more information.
-
 Note that Staged Publishing does not work with Changesets at the moment, so it's recommended to use Trusted Publishing instead for now. Check out [its docs](https://docs.npmjs.com/trusted-publishers) for more information to set it up.
 
 Also, in contrary to npm's [workflow recommendation](https://docs.npmjs.com/trusted-publishers#step-2-configure-your-cicd-workflow), make sure the `id-token: write` is only set on the job that needs to publish. As such, consider splitting the build, test, publish flows etc into separate jobs. Here's an example setup with Changesets:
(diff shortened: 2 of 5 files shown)


docs/set-up-npm-auth.md
```diff
@@ -0,0 +1,144 @@
...
+# For @foo/* packages, publish to a custom registry
+@foo:registry=https://my-registry.com/
+//my-registry.com/:_authToken=${NODE_FOO_AUTH_TOKEN}
+
+# For @bar/* packages, publish to the GitHub Package Registry
+@bar:registry=https://npm.pkg.github.com/
+//npm.pkg.github.com/:_authToken=${GITHUB_TOKEN}
+```
+
+### Package Managers Edge Cases
```
Andarist, line 161: I think a section like this is enough of a reason to go with https://github.com/changesets/action/pull/658 to, at least, ease common scenarios.

bluwy: These aren't common though, except yarn since it literally needs its own config and documentation, but then it's also a simple yarn config that you can create manually which isn't hard.

src/index.ts
```diff
@@ -82,51 +80,6 @@ import {
         "No changesets found. Attempting to publish any unpublished packages to npm",
       );
 
-      if (process.env.NPM_TOKEN) {
```
Andarist, line 85: NPM_TOKEN, IIRC, isn't actually a universal env variable (unlike e.g. `NODE_AUTH_TOKEN`). It's worth rechecking that - but if it was our invention, then perhaps we should throw when it's set, to make the migration easier.

bluwy: `NPM_TOKEN` is a fairly common token name I think, but it being passed to the action is probably the only special part. I'm a bit wary of validating this though since custom publish scripts might want to access this? Worst case the npm publish would also fail anyways so this would only be an early error if implemented. So maybe it's fine leaving as is.

Andarist: ok, let be it

docs/set-up-npm-auth.md
```diff
@@ -0,0 +1,158 @@
...
+          node-version: 24
+          registry-url: https://registry.npmjs.org/ # set this option to set up npm authentication
+      - run: npm install
+      - run: npm build
+      - run: npm publish
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }} # pass the token here
+```
+
+Internally, `actions/setup-node` will set up a configuration like below in `~/.npmrc` (in the home directory):
```
Andarist: as we've learned, it actually won't be in `~/.npmrc`:
https://github.com/actions/setup-node/blob/32f57ac043e003bc1b07edcb2b9102bbbf3e92f1/src/authutil.ts#L48

we probably should vaguely mention that, or at least not mention `~/.npmrc` as that sets up the wrong expectation

README.md
```diff
@@ -166,6 +143,7 @@ jobs:
         uses: actions/setup-node@v6
         with:
           node-version: 26
+          registry-url: https://registry.npmjs.org/
```
Andarist: given this is the default registry... i'd add a comment here

```suggestion
          registry-url: https://registry.npmjs.org/ # makes the action to setup auth line in the config file
```

docs/set-up-npm-auth.md
bluwy: Ah strange I thought it was referring to home before. I'll update this

docs/set-up-npm-auth.md
```diff
@@ -0,0 +1,165 @@
...
+## Recommended Setup
+
+It is recommended by npm to use [Trusted Publishing](https://docs.npmjs.com/trusted-publishers), or [Staged Publishing](https://docs.npmjs.com/staged-publishing), or both, to securely publish packages from CI.
+
+Note that Staged Publishing does not work with Changesets at the moment, so it's recommended to use Trusted Publishing instead for now. Check out [its docs](https://docs.npmjs.com/trusted-publishers) for more information to set it up.
+
+Also, in contrary to npm's [workflow recommendation](https://docs.npmjs.com/trusted-publishers#step-2-configure-your-cicd-workflow), make sure the `id-token: write` is only set on the job that needs to publish. As such, consider splitting the build, test, publish flows etc into separate jobs. Here's an example setup with Changesets:
+
+```yaml
+# .github/workflows/publish.yml
```
Andarist, line 14: shouldnt this workflow also include `/select-mode` and `/version` jobs?

Andarist (approved):

docs/set-up-npm-auth.md
bluwy: I was thinking of only focusing on the publish setup, but yeah probably better to be complete here otherwise people would simply copy and paste this.

PR #697, opened 2026-07-21
Allow custom publish scripts to complete without a Changesets output file

Andarist:


bluwy (approved):
Was this fixing an issue you were facing?

beeequeue:
it's something i found: https://discord.com/channels/1499011163705573470/1528323111894909029/1528365353732673607

could we always set `CHANGESETS_OUTPUT` when we call the custom scripts?

Andarist:
> could we always set CHANGESETS_OUTPUT when we call the custom scripts?

we could - but a custom script can delegate to Changesets and that can create this file. If we’d pre-create it, it would be harder to distinguish between “there was no output” and “the script is output-unaware”

src/run.test.ts
```diff
@@ -20,6 +21,10 @@ vi.mock("@actions/github", () => ({
     graphql: mockedGraphql,
   }),
 }));
+vi.mock("@actions/core", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("@actions/core")>()),
+  warning: vi.fn(),
+}));
```
beeequeue, line 27: we could do this instead, which is cleaner imo

```ts
vi.mock("@actions/core", { spy: true })
const mockedWarning = vi.mocked(core.warning)
```

Andarist: Nice, i’ll update this

PR #699, opened 2026-07-22
Validate that projects use Changesets CLI v3 and direct Changesets CLI v2 users to `changesets/action@v1`

Andarist:


src/utils.ts
```diff
@@ -154,6 +156,55 @@ export function throwOnRenamedInputs(renames: Record<string, string>) {
...
+      throw new Error(changesetsCliCompatibilityError);
+    }
+  }
+
+  let cliPackageJson;
+
+  try {
+    cliPackageJson = require(
+      require.resolve("@changesets/cli/package.json", { paths: [cwd] }),
+    );
```
beeequeue, line 195: why not `import(path, { with: { type: "json" } })`

src/version/index.ts
```diff
@@ -29,6 +33,7 @@ async function main() {
 
   // If the user needs to change the cwd, set `working-directory` in the step instead
   const cwd = process.cwd();
+  await validateChangesetsCliVersion(cwd);
```
bluwy: Should we run this at the start of every `main` instead? Or directly call the function in the top-level try-catch (that calls `main`)?

That way our pre-checks are in a consistent spot.

src/utils.ts
bluwy: I assume it's because the `cwd` could be different

Andarist: yep, that's why we continue using `require` like this here - and in the other place

src/version/index.ts
Andarist: > That way our pre-checks are in a consistent spot.

ye, good suggestion - I changed it accordingly

bluwy (approved):

PR #705, opened 2026-08-03
Fixed a regression making it impossible to sync new versions of versioning pull requests with base branch updates

Andarist:
fixes a regression from https://github.com/changesets/action/pull/675

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "@actions/exec": "^3.0.0",
     "@actions/github": "^9.1.1",
     "@changesets/get-release-plan": "^5.0.0-next.7",
-    "@changesets/ghcommit": "^3.0.0-next.1",
+    "@changesets/ghcommit": "^3.0.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -24,7 +24,7 @@
     "@actions/exec": "^3.0.0",
     "@actions/github": "^9.1.1",
     "@changesets/get-release-plan": "^5.0.0-next.7",
-    "@changesets/ghcommit": "^3.0.0-next.1",
+    "@changesets/ghcommit": "^3.0.0",
```
Andarist, line 27: this is just a driveby update

src/github.ts
```diff
@@ -253,6 +253,7 @@ export class GitHub {
         base: {
           commit: context.sha,
         },
+        force: true,
```
Andarist, line 256: this is the core fix

bluwy (approved):
Don't know why I had removed that. Can you shorten the PR title before merging?

PR #718, opened 2026-08-12
Add back cwd input

### LINKED ISSUE #717: [v2] Nested Changesets projects cannot be used: working-directory is invalid on uses steps
# [v2] Nested Changesets projects cannot be used: `working-directory` is invalid on `uses` steps

## Summary

`changesets/action@v2.0.0` removed the `cwd` input in #659. The associated changeset says to use the step-level `working-directory` option instead.

However, GitHub Actions only supports `working-directory` for `run` steps, not `uses` steps. This leaves no supported way to run the v2 action when the Changesets project root is a subdirectory of the repository.

The action then uses `process.cwd()` as its project directory:

https://github.com/changesets/action/blob/v2.0.0/src/index.ts#L13-L16

## Minimal repository

```text
.
├── .github/workflows/release.yml
├── packages/example/package.json
└── tools/release
    ├── .changeset/config.json
    ├── .changeset/fuzzy-cats-smile.md
    └── package.json
```

`tools/release/package.json` is a private workspace root with `@changesets/cli@^3.0.0` and `../../packages/example` as a workspace.

The Changesets CLI works from that directory and calculates the expected release:

```console
$ cd tools/release
$ ./node_modules/.bin/changeset status --output status.json
🦋 changeset v3.0.0

# status.json contains example-package 1.0.0 -> 1.0.1
```

## Reproduction A: documented replacement is invalid workflow syntax

```yaml
- run: npm install --ignore-scripts
  working-directory: tools/release

- uses: changesets/action@v2.0.0
  working-directory: tools/release
```

Strict workflow validation fails:

```console
$ act --strict --validate --dryrun -W .github/workflows/release.yml
Unknown Property working-directory
```

GitHub documents `jobs.<job_id>.steps[*].working-directory` as an option for `run` steps:

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepsworking-directory

## Reproduction B: schema-valid workflow runs the action at repository root

Removing `working-directory` from the `uses` step makes the workflow valid:

```yaml
- run: npm install --ignore-scripts
  working-directory: tools/release

- uses: changesets/action@v2.0.0
```

Running it produces:

```console
$ act workflow_dispatch --bind --container-architecture linux/amd64 \
    -P ubuntu-latest=catthehacker/ubuntu:act-latest \
    -W .github/workflows/release-root.yml

::error::Error: No package.json could be found upwards from directory <repository-root>
```

The nested install succeeds before the action is invoked. The error occurs because the action uses the repository root as `cwd`, where this fixture intentionally has no `package.json`.

## Expected behavior

There should be a supported way to specify a nested Changesets project root, as there was with the v1 `cwd` input.

Restoring a `cwd` input (or adding a `project-directory` input) and resolving it against `github.workspace` would address this. A regression test should cover a repository whose `.changeset`, package manager manifest, and Changesets CLI installation are all below the repository root.

## Versions

- `changesets/action`: `v2.0.0`
- `@changesets/cli`: `3.0.0`
- `act`: `0.2.89`
- Node: `24.19.0`

## Related history

- #133 / #265 originally established the need for nested-project support.
- #659 removed `cwd` and introduced this regression.

bluwy:
close #717


beeequeue:
can we not just base it on `working-directory` (i.e. the CWD) rather than a duplicate option?

Andarist:
@beeequeue the issue mentions that `working-directory` directory doesn't work on steps using `uses`. This is... peculiar but well :shrug: And `cwd` input does sound better than a lengthy `working-directory` alternative

pack/action.yml
```diff
@@ -7,6 +7,9 @@ inputs:
   publish-plan-artifact-id:
     description: "Artifact id for a publish plan generated by the select-mode subaction"
     required: false
+  cwd:
+    description: "The working directory to execute Changesets in. Defaults to the root of the repository."
```
Andarist, line 11: Technically, it probably defaults to the current workspace dir or smth - but it's probably fine to assume the root of the repository here

.changeset/silent-schools-bathe.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/action": minor
```
Andarist, line 2: One could argue it should be a patch - im fine with either

Andarist (approved):

bluwy:
> And `cwd` input does sound better than a lengthy `working-directory` alternative

Yeah I was also thinking of re-using `working-directory` as the name, but it's probably easier for now to use the same name we had before.

.changeset/silent-schools-bathe.md
bluwy: Yeah I'm also fine with either but I'll merge as minor for now, action versions are usually more lenient

PR #727, opened 2026-08-24
feat(version): configurable pr-body

furcan:
## Context

There are some open discussions around it but no outcome or decisions.

Changesets as developer-intent release management tool can be used/in-use broader then only NPM Packages.
if `package.json` present, and SemVer is a preferred version management for any project, this tool (from CLI to Action) can be used.

The current PR description with pre-state, max content length etc explicitly targeting NPM and Publishing to an NPM artifactory types of wordings.

## Changes

What this PR does is 
- respecting entire logic as is, not creates any regression 
- introducing an optional `pr-body` input, as following the same naming convention that is picked since v2.
- allowing consumers to use `{PACKAGES_INFO}` special identifier in anywhere they want to use in their custom PR body text. `THIS_AND_THAT \n{PACKAGES_INFO}` or just `WHATEVER`

## Additional Code Changes
I can revert these if owners/maintainers don't likes
- `isObject` util moved from `run.ts` to `utils.ts`, extended with array check
- `isString` util added for new `prBody` to allow explicitly only `string` when runtime 
- non-executable code(types) moved to `run.types.ts` file for the touched functions => `getVersionPrBody` + `runVersion`
    - to keep the file simple and readable and by the intent, its runtime code, not build time. 

## Whats remaining

- Get review from owners
- Document this addition on READMEs and related workflows inputs(if any)
- add a changesets entry as minor for this `action`


furcan:
CC: @Andarist 

Thank you in advance 🙇

chj2dctmcr-debug (approved):

chj2dctmcr-debug (approved):

chj2dctmcr-debug (approved):

PR #730, opened 2026-08-26
Improve log messages

bluwy:
Did a subjective pass and tried to improve/make more consistent of the log messages


src/index.ts
```diff
@@ -113,12 +113,11 @@ async function main() {
               : ""
           }`,
         );
-        process.exit(result.exitCode);
```
bluwy, line 116: This isn't needed as I changed to a throw above, which `core.setFailed` will make the exit code 1 by default. But while this might be a different exit code than `result.exitCode`, it doesn't matter in practice as GitHub Actions only has two states: success and failure, which are checked by 0 and non-0 only.

src/run.ts
```diff
@@ -401,7 +401,7 @@ export async function runVersion({
     head: `${context.repo.owner}:${versionBranch}`,
     base: branch,
   });
-  core.info(
+  core.debug(
```
bluwy, line 404: I find this info to bloat the logs in practice, so moved this to debug, which can be viewed when the user re-runs the action in debug mode.

tcka4enckoalexei2014-commits (approved):

tcka4enckoalexei2014-commits (commented):
Go

tcka4enckoalexei2014-commits:
Go

tcka4enckoalexei2014-commits:
1

tcka4enckoalexei2014-commits (approved):

tcka4enckoalexei2014-commits (changes requested):
1

tcka4enckoalexei2014-commits (changes requested):
Go

beeequeue (commented):
i had a todo item to remove the logging of a big json object from the action, but i cant remember if we did that already

bluwy:
The big json object is [this](https://github.com/changesets/action/pull/730/changes#r3859055379) which I move to `core.debug`

beeequeue (commented):
👌

beeequeue (approved):

